PR #16, opened 2026-06-22
fix(storage): neuron_state record-id mismatch broke activation persistence

RobertSigmundsson:
## Summary

`add_neuron` writes the activation-state record as `neuron_state:state_<sid>`, but `get_neuron_state`, `update_neuron_state` and the delete path addressed `neuron_state:<sid>` without the `state_` prefix. `_to_surreal_id` only normalises dashes to underscores; it does not add the prefix. So every `get_neuron_state` missed and returned None, and every update/merge was a silent no-op.

## Impact

The batch helpers delegate to the single-record methods, so it cascades:
- `get_neuron_states_batch` loops `get_neuron_state` and comes back empty.
- `update_neuron_states_batch` loops `update_neuron_state` and persists nothing.

`recall -> ReinforcementManager.reinforce` therefore read empty states and wrote nothing back, so `access_frequency` / `last_activated` / `activation_level` stayed at 0 for every neuron. The whole activation -> decay -> tiering -> consolidation loop was effectively dormant, which is invisible until you go looking for why nothing ages.

## Fix

Three call sites now address `neuron_state:state_<sid>`, matching the writer.

## Type of Change
- [x] Bug fix (non-breaking)

## Testing
- [x] Validated against a live brain with an isolated throwaway record: old key misses, fixed key hits; get + get_batch return the state; update_batch persists `af 0 -> 1` and `last_activated`.

## CHANGELOG (### Fixed)
- `neuron_state` reads/updates/deletes now use the `state_<sid>` record id written by `add_neuron` (they were missing the `state_` prefix, so activation state never persisted and the reinforcement loop was a silent no-op).

Context: #15


acidkill:
Hi Robert — the atomic batch is all in: #17 through #27 are merged. Thank you, genuinely, this was an exceptional set of contributions.

I'm closing this one (#16) on scope. Unlike the others, this branch isn't the "three call sites" fix its title describes — it's a 10-commit integration branch that bundles a superset of #18–#27 together, plus files that belong to your deployment rather than upstream (e.g. `ruflo_crypto.py` and the Docker extras change). Since the individual PRs are already merged cleanly, folding this in would just duplicate them and pull in stack-specific code.

One important thing it leaves behind: the actual neuron_state fix — addressing the state record as `neuron_state:state_<sid>` to match the writer in `add_neuron` — isn't in any of the standalone PRs. It's a real and serious one (the whole activation→decay→tiering→consolidation loop is dormant without it). Could you send it as its own small PR like the rest? I'll take it immediately.

I'll follow up with you separately on the deployment-specific files. Thanks again — looking forward to the neuron_state PR.

RobertSigmundsson:
Closing the loop on this, five weeks late: the `neuron_state` fix you asked for was already in your own `80a47a1` the same afternoon — shipped in v2.4.0 — so there was never a PR for me to send. Nothing outstanding on my side here; thanks for folding it in that fast.

PR #47, opened 2026-07-07
fix: test-config isolation + web UI diversity/tool-stats bugs

acidkill:
## Problem

Running the dev test suite (`pytest tests/`) on a machine with a live
Surreal-Memory station **silently rewrote the real `~/.surrealmemory/config.toml`**,
flipping `current_brain` to `my-brain.v2`. That value then survived reboots (it is
a file) and, via the config → `store.initialize()` auto-create loop, kept
resurrecting a `my-brain.v2` brain — producing a persistent split-brain between
the MCP (env-pinned to `default`) and the CLI (reading the clobbered `config.toml`).

## Root cause (bisected)

`tests/unit/test_mem0_sync_handler.py::TestSwitchBrainValidation::test_valid_brain_name`:

```python
config = UnifiedConfig()            # data_dir defaults to get_surrealmemory_dir() = ~/.surrealmemory
config.switch_brain("my-brain.v2")  # switch_brain() persists via config.save()
```

`UnifiedConfig()` with no `data_dir` targets the **real** `~/.surrealmemory`, and
`switch_brain()` calls `config.save()`, so the hard-coded `"my-brain.v2"` was
written straight into the developer's live station config. Confirmed by bisecting
the whole suite: this one file flips `~/.surrealmemory/config.toml` `current_brain`
to `my-brain.v2`; every other test leaves it alone.

## Fix

Isolate the three `TestSwitchBrainValidation` cases with `data_dir=tmp_path` so
`switch_brain()`/`save()` write to a temp config instead of the real station.

## Verification

- Before: set `current_brain = "default"` → run full suite → reverts to `my-brain.v2`.
- After: set `current_brain = "default"` → run full suite → **stays `default`**.
- Suite: 5805 passed, 4 failed (pre-existing SurrealDB integration/xdist env
  failures, unrelated), no new regressions.

### DIFF
--- a/src/surreal_memory/engine/diagnostics.py
+++ b/src/surreal_memory/engine/diagnostics.py
@@ -143,9 +143,22 @@ def _build_dynamic_action(
             f"(current: {ratio:.1f} synapses/neuron, target: 3.0+)."
         )
     elif component == "diversity":
+        # Diversity is Shannon entropy over synapse types (distribution balance),
+        # normalised against _EXPECTED_SYNAPSE_TYPES common types. A brain can use
+        # MORE than that baseline yet still score low when the mix is skewed, so
+        # never claim "{used} of {expected}" when used >= expected (that read as
+        # the contradictory "16 of 8 expected synapse types used").
+        expected = DiagnosticsEngine._EXPECTED_SYNAPSE_TYPES
+        if types_used < expected:
+            return (
+                f"Use varied memory types — only {types_used} of {expected} common "
+                "synapse types used. Try: 'X caused Y', 'after A then B', "
+                "'X is related to Y'."
+            )
         return (
-            f"Use varied memory types — only {types_used} of 8 expected synapse types used. "
-            "Try: 'X caused Y', 'after A then B', 'X is related to Y'."
+            f"Balance your synapse types — {types_used} types are in use but "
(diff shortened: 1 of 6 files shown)


acidkill:
### Added: two web UI bug fixes (per request)

**Health · diversity card** — the recommendation read the logically-contradictory *"only 16 of 8 expected synapse types used"* (hard-coded 8). Diversity is Shannon entropy over synapse types (distribution balance), so a brain can use more than the expected baseline yet still score low when skewed. The message now branches: a count message (`N of {expected} common types`) only when `N < expected`, otherwise a balance message — no contradiction.

**Tool Stats · Success Rate = NaN%, Avg Duration = NaN** — the SurrealDB `get_tool_stats` `top_tools` entries carried only `tool_name/server_name/count`, while the web UI reads `success_rate` and `avg_duration_ms` per tool (the SQLite backend already returned them). Added `math::mean(duration_ms)` + a per-`(tool, server)` success count (`math::sum(bool)` returns 0 on SurrealDB, so successes are counted via a separate `success = true` GROUP BY), mirrored in the daily period rows.

Verified against live station data: `Bash 192× / 100% / 7748ms`, `smem_remember 16× / 94% / 13700ms`, `smem_recall 12× / 83% / 12086ms` — **0 NaN rows**. New/updated unit tests cover both.

PR #57, opened 2026-07-11
fix(extraction): clause-boundary-aware bigrams + Polish stop-words + auto-mode language-set hygiene

RobertSigmundsson:
Hi Toni!

Following straight on from #56 — this one has been sitting ready on our side for a little while and is, I think, the most broadly useful thing our swarm's production usage has surfaced so far: four small, language-agnostic fixes to auto-mode keyword/concept extraction. Between them they stop the extractor manufacturing junk "concept" bigrams and — the sneaky one — stop it silently deleting real words. On our brain this class of noise accounted for the majority of short junk neurons before we fixed it, so the recall-quality payoff is real and immediate for anyone running auto-mode extraction, whatever their language mix.

Rebased onto current `main` (post-#56); all four commits apply cleanly.

## Summary

The "auto" language path in `extraction/keywords.py` produced junk concept keywords and, worse, silently dropped real words. Four focused fixes:

1. bigrams were formed across sentence/clause boundaries (punctuation was invisible to the position-gap check);
2. the em-dash was not treated as a clause boundary, so one junk-pairing class survived;
3. there was no Polish stop-word set, so bare Polish function words survived as keywords;
4. the Vietnamese stop-word set (ASCII-only entries) was deleting real English/Polish words (`ai`, `anh`, `cho`, `em`, `khi`, `ra`, …) from **every** auto-mode extraction — most notably the word "AI" itself.

All four are quality improvements to the shared extraction path; the explicit per-language branches (`vi`, and the new `pl`) are only reached by explicit-language callers and are unaffected where noted.

## Changes

- **Clause-boundary-aware bigram formation.** Bigrams now require both words to sit in the same clause (split on `.,;:!?\n\r`) **and** a tightened position gap `<= 2` (was `<= 3`), so `"Redis failed. Query timeout occurred."` no longer yields `"failed query"`. Useful stopword-bridged pairs (`query [the] database`) survive; unigram position-decay weighting is untouched (verified bit-for-bit identical on an English control).
- **Em-dash as a clause boundary.** Added to the clause-boundary set; the plain hyphen stays excluded so mid-word compounds (`write-gate` → `write gate`) still work.
- **Polish stop-word set (`STOP_WORDS_PL`).** Conjunctions, prepositions, pronouns, auxiliary verb forms, functional numerals/adverbs — both diacritic and ASCII forms — plus a small conversational set. Folded into the combined auto-mode `STOP_WORDS`, with an explicit `pl` branch in `_get_stop_words`. (`pod` is deliberately excluded — it collides with the container "pod".)
- **Vietnamese stop-words removed from the auto-mode combined set.** `STOP_WORDS_VI` no longer contributes to the auto path — its ASCII-only entries were silently deleting real English/Polish words from every extraction (confirmed live: "AI" and "an em dash" vanished from keywords). The explicit `vi` branch is untouched, so genuine Vietnamese callers are unaffected.

## The four commits

| # | Commit | Rationale |
|---|---|---|
| 1 | `fix(extraction): add Polish stop-word filtering (STOP_WORDS_PL)` | Bare Polish function words (`przez`, `dotyczy`, `ten`, `zostały`, …) survived as concept keywords — the auto branch had no Polish set, and only accidental English-list overlap filtered a few. |
| 2 | `fix(extraction): clause-boundary-aware bigram formation` | Bigram formation only checked the raw word-position gap, so pairs formed across sentence boundaries. Same-clause requirement + gap `<= 2` cuts the junk class while keeping legitimate pairs. |
| 3 | `fix(extraction): treat em-dash as a clause boundary too` | Residual gap caught by a fresh-process behavioural proof: `"keywords.py — bigramy …"` still produced `"py bigramy"`. Same failure class as #2, one hop further. |
| 4 | `fix(extraction): remove Vietnamese stopwords from auto-mode combined set` | The ASCII-only `STOP_WORDS_VI` entries were deleting real words from every auto-mode extraction. |

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- [x] Unit tests added/updated — `test_polish_keywords.py` (new; pins stop-word + clause behaviour incl. the em-dash), `test_vietnamese_removed_from_auto.py` (new), plus updates to `test_vietnamese_keywords.py` and `test_concept_noise_filter.py` (a real-encoder pipeline test built from a live-derived Polish repro sentence).
- [x] Manual testing performed — fresh-process behavioural proofs on live-derived Polish/mixed sentences (documented in the commit bodies).
- Full unit suite on this branch (rebased onto current `main`): **5,842 passed, 33 skipped in ~93 s**, clean process exit. Targeted extraction/keyword/stop-word/bigram selection: **337 passed**. `ruff check` clean on the touched files.

## Checklist

- [x] Code follows the project's style (ruff/mypy per `CONTRIBUTING.md`).
- [x] Self-review performed.
- [x] Tests added that prove the fixes are effective.
- [x] New and existing unit tests pass locally.
- [ ] Documentation updated — n/a (no public API/doc surface changed).
- [ ] `CHANGELOG.md` — proposed entry below; happy to add it as a commit on this branch if you'd like it there rather than in your release pass.

> **Auto-mode keyword/concept extraction no longer forms cross-clause bigrams or drops real words.** Bigrams now require both words in the same clause (split on `.,;:!?\n\r` and the em-dash) with a tightened position gap (`<= 2`). A Polish stop-word set (`STOP_WORDS_PL`, diacritic + ASCII) is added so bare Polish function words stop surviving as keywords. The Vietnamese stop-word set no longer contributes to the auto-mode combined set — its ASCII-only entries (`ai`, `anh`, `cho`, `em`, `khi`, `ra`, …) were silently deleting real English/Polish words from every auto-mode extraction. Explicit-language branches (`vi`, `pl`) are unaffected.

Robert


acidkill:
Thanks Robert — a genuinely valuable batch, and the care shows. I gave it a close read (plus independent checks of the "AI"-survives case and the gap≤2 behaviour): the clause-boundary + em-dash logic and the Vietnamese-set removal from the auto path are correct and well-tested.

One catch I fixed before merge — I pushed a small commit onto this branch. `STOP_WORDS_PL` had `"ci"`, which collides with **"CI"** (continuous integration), ubiquitous in our own dev-session content. It's exactly the ASCII short-form collision this PR removes for Vietnamese (`ai`/`em`), just reintroduced for Polish: `"Fixed CI lint… CI now passes"` dropped "CI" entirely. The commit removes `"ci"` (with a comment so it doesn't creep back) and adds a regression test.

Left for your Polish eye — lower confidence, so I didn't touch them: `ma` (MA / moving-average), `na` (N/A), `co` (CO), `sa` (S.A.). They're real function words, but the same collision-audit methodology you used for VI would settle them cleanly. Two tiny non-blockers too: the en-dash `–` isn't in the clause-boundary set (only em-dash `—`), and the `write-gate` hyphen-non-boundary behaviour has no dedicated test. Neither blocks. CI is green across the board. Thank you again!

PR #58, opened 2026-07-11
fix(store): harden _to_surreal_id against record-id / SurQL injection + single-source the 13 duplicated sanitisers

RobertSigmundsson:
Hi Toni!

This is the one I'd flag as security-sensitive, so I'm sending it as a ready-to-merge fix rather than an issue — no point leaving a window open between describing a breakout and closing it. Nothing exotic, and entirely in keeping with the hardening you're already doing yourself (your `_brain_literal` in 2.7.3 is exactly the same instinct); this just makes that instinct total and single-sourced.

## The problem

`_to_surreal_id` — the function every storage mixin uses to turn a record id into a SurrealDB record name — is `record_id.replace("-", "_")` after stripping the table prefix. It only rewrites hyphens. Quotes, braces, parens, semicolons, whitespace and SurQL/GQL comment markers all pass through untouched, and the result is then **inlined verbatim** into record ids and raw SurQL across the storage layer: `neuron:{sid}`, `UPSERT typed_memory:{sid} CONTENT …`, `DELETE neuron_state:state_{sid}`, and the `eval::gql` fast-path in `_get_path_gql` (`{id:"{sid}"}`).

That is a statement-literal breakout. A hostile `source_id`/`target_id` reaches `get_path` unresolved from the REST `GET /neurons/{source_id}/path` route, and once `eval::gql` is enabled it reaches the GQL parser too.

**Demonstration** (deterministic, no DB required). Feed a breakout id through the current sanitiser:

```python
>>> _to_surreal_id('abc"} ; UPDATE typed_memory SET trust_score=99 RETURN {')
'abc"} ; UPDATE typed_memory SET trust_score=99 RETURN {'   # unchanged
```

Inlined, that yields `UPSERT typed_memory:abc"} ; UPDATE typed_memory SET trust_score=99 RETURN { CONTENT …` — the id has broken out of its own literal into a second statement.

A second, quieter facet: only `store.py`'s copy of `_to_surreal_id` mattered anyway, because **13 storage mixins carry their own duplicate copy** of the function — so even a fix to one copy would leave twelve weak paths. (Verified on current `main`: 13 files define their own `_to_surreal_id`.)

## The fix

Three commits, behaviour-preserving for every legitimate id (UUID4s and content-hashes already live in `[A-Za-z0-9_]` once `-` folds to `_`):

1. **Harden `_to_surreal_id` to a strict fold.** Every character outside `[A-Za-z0-9_]` maps to `_`, not just the hyphen — a single deny-by-default choke point. The same breakout id now comes out inert:

   ```python
   >>> _to_surreal_id('abc"} ; UPDATE typed_memory SET trust_score=99 RETURN {')
   'abc_____UPDATE_typed_memory_SET_trust_score_99_RETURN__'   # [A-Za-z0-9_] only
   ```

2. **A store-layer `_safe_brain_id` guard.** Brain ids legitimately carry `.` and `-` (e.g. `my-brain.v2`), so they aren't folded — instead they are fail-closed *rejected* if they carry anything that could break out of `brain:{id}` / `device:{brain_id}_{did}` / `UPDATE brain:{id} SET …`. It mirrors your REST `_BRAIN_ID_PATTERN` (`[A-Za-z0-9_.-]`, ≤128), so the guarantee no longer depends on the REST layer having validated first — a nice complement to your own `_brain_literal`, which already applies the same charset discipline to the inline-literal path.

3. **Consolidate the 13 duplicate copies into one shared `_ids.py`.** Every mixin now imports the single hardened implementation, so the guarantee is literally true storage-wide rather than true only where someone remembered to harden the local copy. New sinks introduced since (your `find_neurons_by_ids` and `get_edges_for_neurons`, which also inline ids) already route through it.

One test needed a tweak: your `find_neurons_by_ids` test asserted a hostile id is *dropped*; under the strict fold it is instead *neutralised* to an inert record name (`neuron:bad__DELETE_neuron`), which reaches the same safety end — no breakout char survives — so I adjusted the assertion to check the folded, inert form rather than absence.

## Type of Change

- [x] Bug fix (security; non-breaking for legitimate ids)

## Testing

- [x] Unit tests added — `test_id_sanitizer_single_source.py` pins the single-source invariant and the fold, and asserts no mixin retains a private weak copy.
- [x] Existing tests pass, with the one adjusted assertion noted above.
- On our side this was additionally validated against a live SurrealDB with `eval` enabled — a 5,014-payload charset fuzz plus 114 hostile payloads through the REST `get_path` route (0 injections, 0 unexpected 5xx, byte-identical DB) — I've kept that harness out of the PR as it's wired to our deployment, but I'm happy to share it if useful.
- Full unit suite on this branch (rebased onto current `main`): **6,183 passed, 33 skipped in ~30 s**, clean exit. `ruff` and `mypy --strict` clean on the touched files.

## Checklist

- [x] Self-review performed.
- [x] Tests added that prove the fix is effective.
- [x] New and existing unit tests pass locally.
- [ ] `CHANGELOG.md` — proposed entry below; happy to fold it into the branch if you'd like it there.

## Proposed `CHANGELOG.md` entry (Unreleased / Security)

> **Hardened `_to_surreal_id` against record-id / SurQL injection.** Ids are inlined into record literals and raw SurQL across the storage layer; the previous sanitiser only replaced `-` with `_`, leaving quotes, braces, semicolons and comment markers intact — a statement-literal breakout reachable via the REST `get_path` route (and `eval::gql` when enabled). Every character outside `[A-Za-z0-9_]` is now folded to `_`, a store-layer `_safe_brain_id` fail-closed guard covers the dot/hyphen-bearing brain-id path, and the 13 duplicated sanitiser copies are consolidated into one shared implementation so the guarantee holds storage-wide. Behaviour-preserving for legitimate ids (UUID4 / content-hash).

Do give it a sceptical read — security patches deserve nothing less. And thank you again for how open and fast this project is; it's a genuine pleasure to harden something built this thoughtfully.

Robert

### DIFF
--- a/src/surreal_memory/storage/surrealdb/_ids.py
+++ b/src/surreal_memory/storage/surrealdb/_ids.py
@@ -1,0 +1,71 @@
+"""Single hardened source for SurrealDB record-id sanitisation (W7.3 / BUG-10).
+
+Every mixin in ``surreal_memory.storage.surrealdb`` that inlines a caller-supplied
+id into a record literal or raw SurQL string MUST import ``_to_surreal_id`` (and,
+for brain ids, ``_safe_brain_id``) from this module instead of defining its own
+copy. Before BUG-10 this logic was duplicated across 13 files; 12 of those copies
+were the pre-BUG-8 unhardened form (``record_id.replace("-", "_")``, with three
+also stripping a table prefix first), so a hostile id reaching any of those
+mixins could break out of the surrounding string/record literal. Consolidating
+to one module makes the "single choke-point, no un-sanitised id reaches the
+engine" guarantee literally true instead of merely aspirational.
+
+This module must not import from ``store``, any mixin, or any other
+``surreal_memory`` package — stdlib only — so it can be imported by every mixin
+(including ``store.py`` itself) with zero risk of a circular import.
+"""
+
+from __future__ import annotations
+
+
+def _to_surreal_id(record_id: str) -> str:
+    """Convert a record ID to a valid SurrealDB record name (``[A-Za-z0-9_]``).
+
+    Strips any existing table prefix (e.g. 'neuron:abc-123' -> 'abc_123')
+    to prevent doubling when the caller later prepends 'neuron:', then maps
+    every character outside ``[A-Za-z0-9_]`` to ``_``.
+
+    SECURITY (deny-by-default injection guard): the result is inlined verbatim
+    into record-id and query strings — ``neuron:{sid}`` (SurQL), the
+    ``DELETE neuron_state:state_{sid}`` statement, and the ``eval::gql``
+    SHORTEST fast-path in ``_get_path_gql`` (``{id:"{sid}"}``). Legitimate ids
+    (UUID4, content-hashes) already live in this charset once '-' is folded to
+    '_'; enforcing it here means a hostile ``source_id``/``target_id`` (e.g.
+    from the REST ``GET /neurons/{source_id}/path`` route, whose value flows
+    unresolved into ``get_path``) can never carry a ``"``, ``}``, ``)`` or
+    whitespace that breaks out of the surrounding string/record literal to
+    inject SurQL/GQL. Do NOT relax this back to a bare ``.replace('-', '_')``:
+    the old form let ``x"}) ...`` reach the GQL parser once eval was enabled.
+    """
+    if ":" in record_id:
+        record_id = record_id.rsplit(":", 1)[1]
+    return "".join(
+        c if ("a" <= c <= "z" or "A" <= c <= "Z" or "0" <= c <= "9" or c == "_") else "_"
+        for c in record_id
+    )
+
+
+def _safe_brain_id(brain_id: str) -> str:
+    """Validate a brain id before it is inlined raw into a record id / SurQL.
+
+    Brain ids legitimately contain '.' and '-' (e.g. 'my-brain.v2'), so unlike
+    neuron ids they are NOT folded through _to_surreal_id. This is the
+    store-layer choke point that makes the "no un-sanitised id reaches the
+    engine" guarantee literally true: it fail-closed REJECTS any id carrying a
+    quote / brace / paren / semicolon / whitespace / backtick / control char
+    that could break out of the ``brain:{id}`` / ``device:{brain_id}_{did}``
+    record literal or the raw ``UPDATE brain:{id} SET ...`` statement. Mirrors
+    the REST ``_BRAIN_ID_PATTERN`` (``[A-Za-z0-9_.-]``, <=128) so the guarantee
+    no longer depends on the REST layer having validated first.
+    """
+    if (
+        not isinstance(brain_id, str)
+        or not brain_id
+        or len(brain_id) > 128
+        or any(
+            not ("a" <= c <= "z" or "A" <= c <= "Z" or "0" <= c <= "9" or c in "_.-")
+            for c in brain_id
+        )
+    ):
+        raise ValueError(f"unsafe brain id: {brain_id!r}")
+    return brain_id
--- a/src/surreal_memory/storage/surrealdb/store.py
+++ b/src/surreal_memory/storage/surrealdb/store.py
@@ -21,6 +21,7 @@ from surreal_memory.core.fiber import Fiber
 from surreal_memory.core.neuron import Neuron, NeuronState, NeuronType
 from surreal_memory.core.synapse import Direction, Synapse, SynapseType
 from surreal_memory.storage.base import NeuralStorage
+from surreal_memory.storage.surrealdb._ids import _safe_brain_id, _to_surreal_id
 from surreal_memory.storage.surrealdb.activity import SurrealDBActivityMixin
 from surreal_memory.storage.surrealdb.alerts import SurrealDBAlertsMixin
 from surreal_memory.storage.surrealdb.cognitive import SurrealDBCognitiveMixin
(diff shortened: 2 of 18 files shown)


src/surreal_memory/storage/surrealdb/_ids.py
```diff
@@ -0,0 +1,71 @@
...
+    from the REST ``GET /neurons/{source_id}/path`` route, whose value flows
+    unresolved into ``get_path``) can never carry a ``"``, ``}``, ``)`` or
+    whitespace that breaks out of the surrounding string/record literal to
+    inject SurQL/GQL. Do NOT relax this back to a bare ``.replace('-', '_')``:
+    the old form let ``x"}) ...`` reach the GQL parser once eval was enabled.
+    """
+    if ":" in record_id:
+        record_id = record_id.rsplit(":", 1)[1]
+    return "".join(
+        c if ("a" <= c <= "z" or "A" <= c <= "Z" or "0" <= c <= "9" or c == "_") else "_"
```
RobertSigmundsson, line 43: This is the whole fix in one line: every char outside `[A-Za-z0-9_]` folds to `_`, versus the previous `record_id.replace("-", "_")` which only touched hyphens. The output alphabet is a strict subset of valid unquoted record-name characters, so nothing — quote, brace, semicolon, whitespace, comment marker — can survive to break out of an inlined literal. Legitimate ids (UUID4, content-hashes) are unaffected beyond the existing `-`→`_` folding.

src/surreal_memory/storage/surrealdb/_ids.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+
+def _safe_brain_id(brain_id: str) -> str:
+    """Validate a brain id before it is inlined raw into a record id / SurQL.
+
+    Brain ids legitimately contain '.' and '-' (e.g. 'my-brain.v2'), so unlike
+    neuron ids they are NOT folded through _to_surreal_id. This is the
+    store-layer choke point that makes the "no un-sanitised id reaches the
+    engine" guarantee literally true: it fail-closed REJECTS any id carrying a
```
RobertSigmundsson, line 54: Brain ids are deliberately **rejected** rather than folded, because unlike neuron ids they legitimately carry `.` and `-` (e.g. `my-brain.v2`) and folding would corrupt them. The charset here is an exact parity match with your existing REST `_BRAIN_ID_PATTERN` (`[A-Za-z0-9_.-]`, ≤128), so this adds no new policy — it just enforces the same rule at the store layer, so the "no un-sanitised id reaches the engine" guarantee no longer depends on the REST route having validated first. Complements your `_brain_literal` nicely.

src/surreal_memory/storage/surrealdb/store.py
```diff
@@ -465,7 +455,10 @@ def set_brain(self, brain_id: str) -> None:
     def _get_brain_id(self) -> str:
         if self._current_brain_id is None:
             raise ValueError("No brain context set. Call set_brain() first.")
-        return self._current_brain_id
+        # Store-layer choke point: the returned id is inlined raw into
+        # ``device:{brain_id}_{did}`` record literals (register/get/update/delete
+        # device) — fail-closed reject a hostile id so it can never break out.
+        return _safe_brain_id(self._current_brain_id)
```
RobertSigmundsson, line 461: Applying the guard here (rather than only at the REST boundary) is what makes it cover every device-literal sink — `device:{brain_id}_{did}` in register/get/update/delete — from a single place. `save_brain` gets the same treatment for the `brain:{id}` / `UPDATE brain:{id}` sinks. Not a live vector today (REST already validates), so this is defence-in-depth plus making the docstring's guarantee literally true.

tests/unit/test_dashboard_perf_queries.py
```diff
@@ -106,7 +106,11 @@ async def test_empty_and_unsafe_ids_are_dropped(self):
         # An injection-shaped id must not reach the query string.
         await st.find_neurons_by_ids(["bad; DELETE neuron", "ok-1"])
         (sql,) = _queries(conn)
-        assert "DELETE" not in sql
+        # Fork semantics: the hardened single-source _to_surreal_id FOLDS every
+        # char outside [A-Za-z0-9_] to "_", so the hostile id becomes an inert
+        # record id instead of being dropped — no breakout char can survive.
+        assert "neuron:bad__DELETE_neuron" in sql
```
RobertSigmundsson, line 112: Heads-up on the one test change: your original assertion here was `"DELETE" not in sql` — it expected the hostile id to be **dropped**. Under the hardened single-source `_to_surreal_id` the id is instead **folded** to an inert record name (`neuron:bad__DELETE_neuron`), so the word "DELETE" appears inside the id but no breakout character survives. Both reach the same safety outcome; I adjusted the assertion to check the folded, inert form and the absence of any breakout char (`;`, `"`, `'`) rather than absence of the substring.

acidkill:
Thank you — I gave this the sceptical read a security patch deserves (an adversarial pass plus an independent read of `_ids.py`). It holds up completely: the deny-by-default fold, `_safe_brain_id` genuinely fail-closed with the 128-char cap enforced *before* inlining, and tests that prove the invariant structurally — object identity across all 13 modules plus breakout fuzzing, not just behaviour. The store-layer guard also closes a real gap: `consolidation.py` reads `brain_id` straight from a URL path param with no validation. Lovely work, and I appreciate you hardening your own earlier #23.

The one red test (`test_includes_all_14_checks`) wasn't your code — it's a pre-existing xdist isolation flake. The test left `_check_config_freshness` unmocked, so it reads the shared `~/.surrealmemory/config.toml`; under `pytest-xdist` another test touching that file flips `passed` 17→16, and your two new test files just shifted the scheduling enough to expose it. I pushed a one-line mock (like the other 16 checks) plus a ruff-format on the sanitizer test — verified the full suite passes under xdist, and CI is now green.

Two follow-ups I'll track separately (neither blocks this merge): two duplicate weak sanitizers slipped the consolidation — `migrations.py:_sanitize_id` and an inline `.replace("-","_")` in `tool_events.py` — both unreachable today (bound `RecordID` / server-generated uuid4), but they nick the "single choke-point" guarantee; and `find_neurons_by_ids` / `get_neurons_batch` do direct record-id selects without a `WHERE brain_id` scope. Thanks again — exactly the kind of contribution that makes the project better.

PR #59, opened 2026-07-11
fix(surrealdb): reconnect on a dropped transport, not just on 401

RobertSigmundsson:
Hi Toni!

This one's a reliability fix we hit in production and I suspect bites anyone running surreal-memory as a long-lived process against a containerised SurrealDB — which, given the SurrealDB-first direction, is most of us. Small diff, clear failure mode.

## The problem

`SurrealDBStorage._query` retries the query once — but only when `_is_auth_error(exc)` matches, i.e. the exception message contains `401`/`unauthorized`. That covers the ~1h root-token TTL case the retry was originally written for. It does **not** cover a connection dropped for any *non-auth* reason: a DB container restart (nightly backup, image upgrade, host reboot) severs the WebSocket, which surfaces as a connection/transport error, not a 401. So `_is_auth_error` returns False, `_query` re-raises, and the cached dead connection is never replaced — it then fails **every** subsequent query, and the whole MCP surface returns `-32000` until the process itself is restarted.

We saw exactly this: our SurrealDB container restarted, the CLI (fresh process) kept working perfectly, while the long-lived MCP server answered `-32000` to every single tool call until we bounced it. The symptom looks like random tool breakage rather than a connection issue, which is what makes it worth closing.

## The fix

Add `_is_connection_error(exc)` — matches `ConnectionError`/`OSError` subclasses, exception type names (`connectionclosed`, `disconnect`, `websocket`, `transport`), and message fragments (`connection closed/reset/refused/lost`, `not connected`, `broken pipe`, `going away`) — and reconnect in `_query` when **either** the auth check **or** the connection check matches. Same `_reauth_lock`, same single retry, same `_reconnect()` path. A genuine query/syntax error still re-raises unchanged, so this never masks a real error behind a needless reconnect.

## Type of Change

- [x] Bug fix (reliability; non-breaking)

## Testing

- [x] Unit tests added — `TestConnectionErrorDetection` (4 cases): stdlib `ConnectionResetError`/`OSError`, WebSocket close messages (`going away`, `websocket connection is closed`, `Connection refused`), a class-name match (`ConnectionClosedError`), and — importantly — that a parse/query error and a `ValueError` are **not** flagged, so no needless reconnect.
- Full unit suite on this branch (based on current `main`): **5,821 passed, 33 skipped in ~36 s**, clean exit. `ruff` and `mypy --strict` clean on the touched file.

## Checklist

- [x] Self-review performed.
- [x] Tests added that prove the fix.
- [x] `CHANGELOG.md` — proposed entry below; happy to fold it into the branch.

> **`SurrealDBStorage` now reconnects after a dropped transport, not only on a 401.** A DB container restart (backup, upgrade, reboot) severs the WebSocket and surfaces as a connection error rather than an auth error, so the previous retry-on-401-only path left the cached dead connection in place — every subsequent query failed and long-lived clients returned `-32000` until the process was restarted. `_query` now also reconnects on connection/transport errors (`_is_connection_error`), using the same single-retry path; genuine query errors are unaffected.

Robert

### DIFF
--- a/src/surreal_memory/storage/surrealdb/store.py
+++ b/src/surreal_memory/storage/surrealdb/store.py
@@ -115,7 +115,14 @@ def _is_connection_error(exc: Exception) -> bool:
     and the whole MCP surface returns -32000 until the process is restarted
     (audit finding S-01).
     """
-    if isinstance(exc, (ConnectionError, ConnectionResetError, OSError)):
+    # OSError covers ConnectionError/ConnectionResetError (both subclasses) and
+    # the aiohttp ClientOSError/ClientConnectorError raised during a restart
+    # window. TimeoutError (== asyncio.TimeoutError since 3.11) is ALSO an OSError
+    # subclass, but a legitimate slow-query timeout under the default HTTP
+    # transport (ClientTimeout(total=30)) is a query outcome, not a dropped
+    # transport — excluding it avoids a needless reconnect+retry that doubles
+    # latency and masks the SDK's own is_timed_out error class.
+    if isinstance(exc, OSError) and not isinstance(exc, TimeoutError):
         return True
     name = type(exc).__name__.lower()
     if any(k in name for k in ("connectionclosed", "disconnect", "websocket", "transport")):
--- a/src/surreal_memory/storage/surrealdb/store.py
+++ b/src/surreal_memory/storage/surrealdb/store.py
@@ -105,6 +105,38 @@ def _brain_literal(brain_id: str) -> str:
     return f'"{brain_id}"'
 
 
+def _is_connection_error(exc: Exception) -> bool:
+    """True if an exception signals a dropped/closed transport rather than an
+    auth or query error.
+
+    A long-lived WebSocket connection severed by a DB container restart (backup,
+    upgrade, reboot) never returns 401, so ``_is_auth_error`` misses it. Without
+    reconnecting on this, the cached dead connection fails EVERY subsequent query
+    and the whole MCP surface returns -32000 until the process is restarted
+    (audit finding S-01).
+    """
+    if isinstance(exc, (ConnectionError, ConnectionResetError, OSError)):
+        return True
+    name = type(exc).__name__.lower()
+    if any(k in name for k in ("connectionclosed", "disconnect", "websocket", "transport")):
+        return True
+    msg = str(exc).lower()
+    return any(
+        s in msg
+        for s in (
+            "connection closed",
+            "connection reset",
+            "connection refused",
+            "connection lost",
+            "not connected",
+            "no connection",
+            "websocket",
+            "broken pipe",
+            "going away",
+        )
+    )
+
+
 def _from_surreal_id(surreal_id: str) -> str:
     """Extract the original ID from a SurrealDB record ID like 'neuron:abc_def'."""
     if ":" in surreal_id:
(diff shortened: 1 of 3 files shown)


src/surreal_memory/storage/surrealdb/store.py
```diff
@@ -115,6 +115,38 @@ def _brain_literal(brain_id: str) -> str:
     return f'"{brain_id}"'
 
 
+def _is_connection_error(exc: Exception) -> bool:
```
RobertSigmundsson, line 108: Kept deliberately broad — a dropped WebSocket surfaces very differently across the SDK / websockets / OS stack (a typed `ConnectionError`, a `ConnectionClosed` class name, or just a `going away` / `broken pipe` string), so this matches on all three: exception subclass, type name, and message fragment. The one thing it must NOT catch is a genuine query/syntax error (that would loop a reconnect pointlessly) — the last test case pins exactly that.

src/surreal_memory/storage/surrealdb/store.py
```diff
@@ -494,11 +526,15 @@ async def _query(self, sql: str, **params: Any) -> list[dict[str, Any]]:
...
+
+        Also reconnects on a dropped transport (WebSocket closed by a DB restart):
+        that surfaces as a connection error, not a 401, so without this the dead
+        cached connection fails every query until the process restarts (S-01).
         """
         try:
             result = await self._ensure_conn().query(sql, params)
         except Exception as exc:
-            if not _is_auth_error(exc):
+            if not (_is_auth_error(exc) or _is_connection_error(exc)):
```
RobertSigmundsson, line 537: The only behavioural change: `if not _is_auth_error(exc)` becomes `if not (_is_auth_error(exc) or _is_connection_error(exc))`. Everything downstream — the `_reauth_lock`, the single `_reconnect()`, the one retry — is your existing 401 path, reused verbatim. So a transport drop now heals exactly the way a token expiry already did, and nothing else moves.

acidkill:
Clean, well-scoped fix. The mechanics are right — single retry, shared `_reauth_lock`, and `_reconnect()` actually replaces the cached connection so subsequent queries recover — and I verified the predicate against the vendored surrealdb 2.0.0 SDK's actual exception types.

One thing I fixed before merge (pushed to the branch): `TimeoutError` (== `asyncio.TimeoutError` since 3.11) is an `OSError` subclass, so `_is_connection_error` was flagging a legitimate slow-query timeout under the default HTTP transport (`ClientTimeout(total=30)`) as a dropped transport — a needless reconnect+retry that doubles latency and hides the SDK's own `is_timed_out`. Given our full-scan queries can run for seconds, that's reachable. The commit narrows the `OSError` branch to exclude `TimeoutError`, adds a regression test, and applies the ruff-format fix. CI is green.

One note for the audit trail: the dead-cached-connection symptom reproduces on the WS transport, but our deploys use HTTP (a fresh `ClientSession` per query, so no cached socket to go stale) — the fix is still correct and genuinely helps anyone on `ws://`. Nice catch on the failure mode; thank you!

acidkill:
Quick follow-up: after #58 merged (it also touched `store.py` + `test_surrealdb_store.py`), this branch conflicted. I rebased it onto current `main` and resolved the conflict — which was purely in the test file, where both sides had appended classes at the end. Kept everything: your `TestConnectionErrorDetection` (+ the TimeoutError regression test) alongside the id-sanitiser test classes #58 added. `store.py` merged cleanly, so the hardening and the reconnect/TimeoutError changes coexist. Force-pushed; CI is green and the conflict is gone. Ready to merge.

acidkill (approved):
great work. thanks!

Issue #64, opened 2026-07-12
Polish stop-words ma/na/co/sa: the N/A and S.A. collisions don't reproduce — propose acronym-preservation over removal


RobertSigmundsson:
Hi Toni — following up on the `ma` / `na` / `co` / `sa` question you left for "the Polish eye" in #57. I ran the same collision audit we used for Vietnamese, with the real extractor, and it points somewhere different from removing them — so an issue rather than a PR.

## What the audit found (live, `language="pl"`)

The tokenizer keeps only letter runs on lowercased text (`re.findall(r"\b[a-zA-ZÀ-ỹ]+...\b")` over `text.lower()`), which changes the picture:

| word | punctuated form | bare acronym | Polish function word (noise if removed) |
|---|---|---|---|
| `na` | **`N/A` does NOT collide** — the `/` splits it into `n` + `a`, both below min-length | `NA` (North America) collides | `na` (on/at) — one of the most frequent PL words |
| `sa` | **`S.A.` does NOT collide** — the `.` splits it into `s` + `a` | `SA` collides (much rarer than `S.A.`) | `są`/`sa` (are) — common |
| `co` | `C/O` does NOT collide | `CO` collides | `co` (what) — common |
| `ma` | (no punctuated form) | `MA` (moving average) collides | `ma` (has) — very frequent PL verb |

So the specific collisions that motivated `na` and `sa` (`N/A`, `S.A.`) don't actually happen — the punctuation fragments them below the min-length guard. Only the **bare uppercase** acronyms collide, and for all four the Polish function word is far more common in real content than the acronym. That's the inverse of the `ci` ↔ `CI` case, where the acronym was ubiquitous and the function word rare.

**Conclusion:** removing `ma`/`na`/`co`/`sa` from `STOP_WORDS_PL` would trade a lot of function-word noise for a niche acronym gain. I don't think it's worth it as a straight removal.

## The better fix: preserve uppercase acronyms

The root cause is that `extract_weighted_keywords` lowercases up front (`_CLAUSE_BOUNDARY.split(text.lower())`), which conflates the acronym `MA` with the function word `ma`. A targeted fix would keep a token that was ALL-CAPS (≥2 chars) in the original text even when its lowercased form is a stop-word — so `MA`, `CO`, `NA`, `SA` (and every other acronym) survive, without letting the Polish function words through.

[...]

PR #65, opened 2026-07-12
fix(extraction): treat the en-dash as a clause boundary too, and pin hyphen non-boundary

RobertSigmundsson:
Hi Toni!

The two non-blockers you flagged on #57, folded into one small change.

## En-dash as a clause boundary
`_CLAUSE_BOUNDARY` had the em-dash (—) but not the en-dash (–, U+2013). Editors and operating systems emit the en-dash just as often for ranges and asides, so the same cross-clause junk bigram the em-dash fix targets (a filename fragment glued onto the next clause) still slipped through when the separator was an en-dash. Added it to the set.

## Hyphen non-boundary — now pinned
You noted the hyphen-non-boundary behaviour had no dedicated test. Added one: a plain ASCII hyphen must NOT split a clause, so compound halves like `cache-aside` still pair as a bigram — kept as an explicit contrast to the en-/em-dash, which do split.

## Type of Change
- [x] Bug fix + regression test (non-breaking)

## Testing
- `test_no_bigram_across_en_dash` and `test_hyphen_is_not_a_clause_boundary` added to `TestClauseBoundaryBigrams`.
- Full unit suite (rebased on `main`): **6,082 passed, 33 skipped in ~30 s** (`-n0`). `ruff` + `mypy` clean. (The literal en-dash in the regex and one test string carry a `# noqa: RUF001` — ruff flags a bare en-dash as confusable with a hyphen, but here we want it literal.)

## Note on the ma/na/co/sa question (separate)
I ran the collision audit you suggested for the four Polish stop-words; it points at a different fix than removal (the `N/A`/`S.A.` collisions don't actually reproduce), so I opened #64 for that discussion rather than bundling it here.

Robert

### DIFF
--- a/src/surreal_memory/extraction/keywords.py
+++ b/src/surreal_memory/extraction/keywords.py
@@ -554,12 +554,13 @@ STOP_WORDS: frozenset[str] = (
 )
 
 # Clause boundary: bigrams never pair words that cross one of these.
-# Em-dash is included — it sets off a parenthetical/separate clause the same as a
-# comma (live-proven: "keywords.py — bigramy" otherwise lets the "py" filename
-# fragment glue onto the next clause's first word). Plain ASCII hyphen is
+# Em-dash (—) and en-dash are both included — each sets off a parenthetical /
+# separate clause the same as a comma (live-proven: "keywords.py — bigramy"
+# otherwise lets the "py" filename fragment glue onto the next clause's first
+# word; editors and OSes emit the en-dash just as often). Plain ASCII hyphen is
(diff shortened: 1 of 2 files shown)


acidkill:
Good catch, and the reasoning holds: editors and OSes emit U+2013 for ranges and asides just as readily as U+2014, so leaving it out left exactly the same cross-clause junk bigram in play. Keeping the ASCII hyphen out of the class is the right call — `write-gate` must still yield `write gate`.

Test pins the behaviour rather than the implementation. Merging.

PR #84, opened 2026-07-17
perf(doc-trainer): skip cross-linking/emotion + inline type in train mode (−42%)

acidkill:
Data-driven fix from cProfile + per-query wall profiling on a 92k-neuron default brain (N=20 chunks). find_neurons at 1.1-1.4 s/q dominated (96% DB I/O).

## Root cause
SDB 3.2.0 planner picks `idx_neuron_type` and filters content post-index — scanning 40k+ type-partition rows per call. Even with `type` as literal (IndexScan), the content `WHERE` is applied as a post-index filter.

## Changes (all gated on `ctx.skip_conflicts` = train mode)
- **SemanticLinkingStep + CrossMemoryLinkStep**: skip (ENRICH consolidation adds cross-links post-train). Removes ~90 find_neurons calls/20 chunks.
- **EmotionStep**: skip (doc knowledge has no emotional valence).
- **find_neurons**: inline `type` as literal (enum-safe) so planner uses `idx_neuron_type` composite (IndexScan vs scan+filter).
- **find_neurons_exact_batch**: keep `content IN` override but inline type literal. N+1 base default was 4-5× slower (each type+content call ≈ 750ms vs one IN-query ≈ 1.3s for N=8 candidates).

## Measurement
On Toni's 92k-neuron default brain, N=20 chunks: **6.7 → 3.85 s/chunk (−42%)**.

## Test plan
- [x] `ruff check` + `ruff format --check` clean
- [x] profiled N=20 on live default brain (data-driven, not bench)
- [ ] CI (ruff/mypy/pytest)

## Verified by
@acidkill

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,56 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.11.0] — Fast bulk doc-training
+
+`smem train` on large documents is dramatically faster on big brains, and the
+encoder now batches its writes instead of issuing one round-trip per neuron /
+synapse. Three layers of work.
+
+### Added
+
+- `SurrealDBStorage.add_synapses_batch` (multi-statement `INSERT RELATION`)
+  plus a base `add_synapses_batch` default with a per-synapse fallback, so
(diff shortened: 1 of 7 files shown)


acidkill:
Replaced by clean branch (rebased on main, no conflicts).

PR #90, opened 2026-07-26
fix(cli): honour SURREAL_MEMORY_BRAIN when --brain is omitted

RobertSigmundsson:
### Problem

`get_storage` has its own env-var fallback for the brain name, but it only fires when the caller passes
`None`. Several CLI commands need the brain name *before* calling `get_storage` (to echo progress), and
they resolve it with the `brain or config.current_brain` pattern. That expression is never `None`, so
`get_storage` skips its own env check and **`SURREAL_MEMORY_BRAIN` is silently ignored** — the command
prints and operates on the config-file brain instead of the requested one.

### Change

Add a single `resolve_brain(brain, config, *, default=None)` helper in `cli/_helpers.py` implementing the
documented priority — **explicit arg > env var > default/config** — and route `brain.py`, `shortcuts.py`
and `tools.py` through it. The helper's docstring states why the `brain or config.current_brain` pattern
must not be used directly.

### Why upstream benefits

`SURREAL_MEMORY_BRAIN` is a documented knob; today it works for some commands and not others, with no
error to indicate which. One helper makes the precedence rule single-sourced and testable.

### Side effect worth noting

`main` currently has a redundant expression at three call sites, e.g. `cli/_helpers.py:90`:

```python
env_brain = os.environ.get("SURREAL_MEMORY_BRAIN") or os.environ.get("SURREAL_MEMORY_BRAIN")
```

— the same lookup on both sides of the `or` (also `cli/commands/brain.py:39` and `:87`). Harmless, but a
clear leftover. Routing through `resolve_brain` removes all three.

## Evidence

| Check | Result |
|---|---|
| `pytest -m "not stress" -n auto` | **6542 passed**, 84 skipped, 1 xfailed (baseline 6535 → **+7**) |

New file: `tests/unit/test_cli_resolve_brain.py` (+59) covering all three precedence branches.

- **Behaviour change by design:** a user who exported `SURREAL_MEMORY_BRAIN` and relied on those specific
  commands ignoring it will now see them honour it. That was the bug.
- No storage-layer changes; backend-agnostic.

### DIFF
--- a/src/surreal_memory/cli/_helpers.py
+++ b/src/surreal_memory/cli/_helpers.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
+import os
 from collections.abc import Coroutine
 from pathlib import Path
 from typing import Any, TypeVar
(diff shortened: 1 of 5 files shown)


acidkill:
This fixes a real bug and removes a copy-paste artefact in the same pass: `os.environ.get("SURREAL_MEMORY_BRAIN") or os.environ.get("SURREAL_MEMORY_BRAIN")` read the same key twice, so the fallback never did anything.

The important part is the invariant in the docstring: `brain or config.current_brain` always yields a non-`None` name, which makes `get_storage` skip its own env lookup. That is the kind of bug that stays invisible until someone actually sets the env var and it is quietly ignored. Routing every call site through one helper is the right fix.

Merging.

PR #91, opened 2026-07-26
feat(embedding): BGE-M3 HTTP provider (1024D, zero-vector guard) + full registration

RobertSigmundsson:
### Problem

`main` already assumes a self-hosted bge-m3 endpoint in several places
(`unified_config.py:124`, `unified_config.py:2521`, `engine/encoder.py:461`,
`engine/reasoning_distiller.py:8` and `:410`) but there is no `bge_m3` provider — you can only reach such
an endpoint through the OpenAI-compatible path, which does not carry BGE-M3's dimension or its
normalisation contract.

### Change

Four commits, one logical addition:

1. `BGEM3Embedding` — posts to a `/embed` HTTP service, returns 1024-D L2-normalised vectors, and
   **raises on empty/failed input instead of fabricating a zero vector**. A fabricated zero vector is
   silently catastrophic for a cosine KNN index: it poisons the neighbourhood without any error.
2. Registration in `unified_config._VALID_PROVIDERS` — without it the effective-config validator
   rejected `bge_m3` even though the factory could build it.
3. Generic wording in the provider docstring (no deployment-specific references).
4. Registration in the embedding **capability probe** — otherwise recall is reported as `degraded`
   even when the provider is working.

Configuration is env-driven (`SURREAL_MEMORY_EMBEDDING_BASE_URL`, `BGE_M3_API_KEY`,
`SURREAL_MEMORY_EMBEDDING_DIMENSION`), consistent with the other providers.

### Why upstream benefits

Self-hosted BGE-M3 is a common way to run this project without a paid embedding API, and the codebase
already talks about it. This makes it a first-class provider, including the two registration points that
are easy to miss (items 2 and 4 were each found only after the provider silently misbehaved).

## Evidence

| Check | Result |
|---|---|
| cherry-pick onto `origin/main` | clean (4 commits) |
| `mypy src/ --ignore-missing-imports` | Success: no issues found in **370** source files (369 + the new module) |

- **No new tests.** The provider is pure HTTP I/O and this branch adds no unit test for it; the suite is
  unchanged at 6535. If upstream wants a mocked-transport test before merging, say so and it will be added.
- Additive only: no existing provider path is modified, so an installation that never sets
  `provider = "bge_m3"` sees no behaviour change.
- Does **not** touch the storage layer or any KNN contract — it plugs into the existing
  `semantic_discovery._create_provider` factory.
- Dimension mismatch caveat: switching an existing brain from a 768/3072-D provider to this one requires
  a re-embed, same as any provider change. Not handled here (and not handled for the other providers either).

### DIFF
--- a/src/surreal_memory/engine/embedding/bge_m3_embedding.py
+++ b/src/surreal_memory/engine/embedding/bge_m3_embedding.py
@@ -1,7 +1,6 @@
 """BGE-M3 embedding provider (HTTP, dense, L2-normalized).
 
-Talks to a self-hosted BGE-M3 FastAPI service (e.g. on vast.ai reached via an
-SSH tunnel at http://127.0.0.1:18100):
+Talks to a self-hosted BGE-M3 FastAPI service exposed on a local endpoint:
 
     POST {base_url}/embed   body {"texts": [...]}   header Authorization: Bearer <key>
     -> {"embeddings": [[... 1024 floats ...], ...]}   (already L2-normalized)
--- a/src/surreal_memory/engine/embedding/bge_m3_embedding.py
+++ b/src/surreal_memory/engine/embedding/bge_m3_embedding.py
@@ -1,0 +1,123 @@
+"""BGE-M3 embedding provider (HTTP, dense, L2-normalized).
+
+Talks to a self-hosted BGE-M3 FastAPI service (e.g. on vast.ai reached via an
+SSH tunnel at http://127.0.0.1:18100):
+
+    POST {base_url}/embed   body {"texts": [...]}   header Authorization: Bearer <key>
+    -> {"embeddings": [[... 1024 floats ...], ...]}   (already L2-normalized)
+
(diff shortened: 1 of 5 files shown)


acidkill:
Reviewed the provider rather than just the wiring. It talks to a different API shape than the OpenAI-compatible path (`POST /embed` with `{"texts": [...]}` versus `/v1/embeddings`), so it is not a duplicate of the existing route, and the registration is isolated — nothing changes for anyone who does not select it.

The zero-vector guard is the part I would have insisted on: returning a fabricated `[0.0] * dim` on failure poisons top-k KNN for every later query, and it does so silently. Raising and letting the caller leave the neuron pending is correct.

One follow-up, no action needed here: this introduces `SURREAL_MEMORY_EMBEDDING_BASE_URL` while the OpenAI-compatible provider reads `SURREAL_MEMORY_EMBEDDING_ENDPOINT`. Two similarly-named variables for the same conceptual thing is a setup where people set the wrong one and get a confusing fallback. Worth unifying (or documenting side by side) before this gets much use.

Merging.

PR #92, opened 2026-07-26
feat(reranker): support Bearer auth and the `score` response field in HttpReranker

RobertSigmundsson:
### Problem

`HttpReranker` sends `Content-Type` only, and parses the response with
`float(r["relevance_score"])`. Two consequences:

1. Any rerank endpoint behind a token (most self-hosted deployments that are not on `localhost`)
   cannot be used at all.
2. Endpoints that return the field as `score` — including BGE-M3's `/rerank` — raise `KeyError`
   instead of reranking.

### Change

- Optional `api_key` constructor argument, falling back to `SURREAL_MEMORY_RERANKER_API_KEY` then
  `BGE_M3_API_KEY`. An empty key sends **no** `Authorization` header, so llamastash / llama.cpp
  deployments that need none are unaffected.
- Response parsing accepts both field names:
  `float(r.get("relevance_score", r.get("score", 0.0)))` — OpenAI-compatible field first, `score` as fallback.

### Why upstream benefits

Two small compatibility gaps that each make a whole class of rerank endpoints unusable. The change is
strictly additive: existing unauthenticated, `relevance_score`-returning endpoints behave exactly as before.

## Evidence

| Check | Result |
|---|---|

- **No new tests** (suite unchanged at 6535). Happy to add a mocked-`urlopen` test covering
  "no key → no header", "key → header", and both response field names if upstream wants it before merge.
- `r.get("score", 0.0)` defaults a missing score to `0.0` rather than raising. That is deliberately
  lenient — a malformed row sorts last instead of killing the whole rerank — but it is a judgement call
  worth a reviewer's opinion.
- The `# noqa: S310` comment was retightened from "fixed local llamastash endpoint" to
  "fixed local rerank endpoint" since the endpoint is no longer llamastash-specific.

### DIFF
--- a/src/surreal_memory/engine/reranker.py
+++ b/src/surreal_memory/engine/reranker.py
@@ -182,26 +182,42 @@ class HttpReranker:
         blend_weight: float = 0.7,
         max_candidates: int = 30,
         timeout: float = 15.0,
+        api_key: str = "",
     ) -> None:
         self._endpoint = endpoint.rstrip("/")
         self._model = model_name
         self._blend_weight = min(max(blend_weight, 0.0), 1.0)
         self._max_candidates = min(max_candidates, 100)
         self._timeout = timeout
+        # Bearer auth for endpoints that require it (e.g. the self-hosted BGE-M3
+        # /rerank service). llamastash/llama.cpp need none — an empty key sends no header.
+        self._api_key = (
+            api_key
+            or os.environ.get("SURREAL_MEMORY_RERANKER_API_KEY", "")
+            or os.environ.get("BGE_M3_API_KEY", "")
+        ).strip()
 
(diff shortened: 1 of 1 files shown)


acidkill:
Narrow and backward compatible — an empty key sends no header, so llamastash and llama.cpp are untouched. Accepting both `relevance_score` and `score` is the pragmatic read of a field that is not actually standardised across rerank services.

Worth noting the interaction with #97, which landed earlier today: recall now reports rerank degradation instead of silently falling back to raw spreading-activation order, and the reason string is sanitised (URLs stripped, length capped) precisely because it travels into the recall response. That mattered while the endpoint was assumed local; with Bearer auth in play it matters more.

Merging.

PR #93, opened 2026-07-26
feat(write-gate): shadow/enforce modes, per-intent auto-capture mode, and gate_decision telemetry

RobertSigmundsson:
### Problem

`WriteGateConfig` on `main` is a single boolean: `enabled`. That forces an all-or-nothing choice.
Turning it on immediately starts rejecting writes with thresholds nobody has measured on their own
corpus; leaving it off means the gate never produces any data you could tune it with. In practice a
global `enforce` false-rejects real conversational turns and summaries, so the gate ends up permanently off.

### Change

**1. Three modes instead of a boolean.**

```python
mode: str = "off"            # off | shadow | enforce — overrides `enabled` when not "off"
auto_capture_mode: str = ""  # "" (inherit) | off | shadow | enforce
```

`shadow` evaluates the gate and records the decision **without rejecting anything**. Backward compatible:
`effective_mode` falls back to the legacy `enabled` bool (`True` → `enforce`) whenever `mode` is `"off"`,
so existing configs behave exactly as today.

**2. Per-intent override.** `auto_capture_mode` applies to `intent=auto` only; summaries and turns keep
`mode`. This is the practically useful combination: **enforce** the gate on junk passive auto-captures
while interactive writes stay in **shadow**.

**3. `gate_decision` telemetry** (`engine/gate_telemetry.py`, new) wired into the paths that actually
write: `mcp/remember_handler.py`, `hooks/stop.py`, `hooks/pre_compact.py`. This is what makes shadow mode
worth anything — you get the accept/reject record you need to pick thresholds before you enforce them.

**4.** Two test commits replace hand-rolled config mocks in the MCP and geo server tests with a real
`WriteGateConfig`, so the mocks cannot drift from the dataclass.

### Why upstream benefits

Shadow mode is the standard way to ship a policy engine: measure first, enforce second. Without it the
write gate is a feature most users will never safely turn on. Full backward compatibility means zero
impact on anyone who does not set `mode`.

## Evidence

| Check | Result |
|---|---|
| cherry-pick onto `origin/main` | clean (4 commits) |
| `mypy src/ --ignore-missing-imports` | Success: no issues found in **370** source files (369 + `gate_telemetry`) |
| `pytest -m "not stress" -n auto` | **6541 passed**, 84 skipped, 1 xfailed (baseline 6535 → **+6**) |

`tests/unit/test_write_gate.py` +70 lines covering `effective_mode` / `effective_auto_mode` resolution
including the legacy-`enabled` fallback.

- Largest of the eight PRs (8 files). Splittable if preferred: commits 1+3+4 (config + test mocks) are
  independent of commit 2 (telemetry + wiring), though shadow mode is not much use without the telemetry.
- Two config knobs where there was one boolean. `effective_mode` / `effective_auto_mode` centralise the
  precedence so no call site re-derives it — worth a look, that is where a subtle bug would live.
- Telemetry writes: `gate_decision` records are persisted. A reviewer should confirm the retention story
  is acceptable — this branch does not add pruning for them.
- Backend-agnostic: no storage-layer schema change.

### DIFF
--- a/src/surreal_memory/engine/gate_telemetry.py
+++ b/src/surreal_memory/engine/gate_telemetry.py
@@ -1,0 +1,65 @@
+"""Write-gate telemetry — append engine-side gate decisions to the
+`gate_decision` table (the same SCHEMALESS table the Hermes plugin writes to),
+so scripts/gate_stats.py sees BOTH the plugin and the engine (smem
+auto-capture / stop-hook) decisions in one report.
+
+Fire-and-forget: logging must never break a write.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+
+def _lang_hint(s: str) -> str:
+    """'cjk' if the text contains CJK / Kana / Hangul, else 'latin' (no lib)."""
+    for ch in s:
+        o = ord(ch)
+        if (
+            0x4E00 <= o <= 0x9FFF
+            or 0x3400 <= o <= 0x4DBF
+            or 0x3040 <= o <= 0x30FF
+            or 0xAC00 <= o <= 0xD7A3
+        ):
+            return "cjk"
+    return "latin"
+
+
+async def log_gate_decision(
+    storage: Any,
+    *,
+    intent: str,
+    accepted: bool,
+    reason: str,
+    score: int | None,
+    mode: str,
+    content: str,
+    agent_id: str = "",
+) -> None:
+    """Best-effort insert of one gate decision (both SHADOW and ENFORCE log)."""
+    try:
+        from surreal_memory.utils.timeutils import utcnow
+
+        conn = storage._ensure_conn()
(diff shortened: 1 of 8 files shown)


acidkill:
The framing convinced me more than the code did: an all-or-nothing gate is one you either never turn on, or turn on once and then turn off after the first false reject. Shadow mode is what makes the thresholds tunable against a real corpus instead of guessed.

Checked the compatibility claim rather than taking it on trust: `mode` defaults to `"off"` and `effective_mode` falls back to the legacy `enabled` bool, so merging this changes behaviour for exactly nobody until someone opts in. That is what makes it safe to land now rather than parking it behind a release.

The per-intent override is the genuinely useful part — enforcing on passive auto-captures while interactive writes stay in shadow is the combination that matches how the noise actually arrives. #94, merged just before this, cuts the same problem from the other end.

Merging.

PR #94, opened 2026-07-26
feat(write-gate): reject machine output and session-notification noise (Gate 1b)

RobertSigmundsson:
### Problem

Passive auto-capture stores a lot of harness output as if it were knowledge. Measured over one day of
`gate_decision` false-accepts on a live brain: **180×** `"Session activity:"`, **30×** `<summary>`,
**20×** `"Background command"`, **19×** `<status>`, **12×** `task-notification`, plus `tool-use-id`,
`*-caveat>` and raw shell lines (`ps -o …`, `echo …`, `curl -…`). None of it is a memory. All of it
passes the length and quality-score gates because it is long, structured, and full of distinct tokens.

### Change

A new `Gate 1b` in `score_memory`, running before the scoring heuristics:

```python
def _is_machine_noise(content: str) -> bool:
    """Two signals: (a) a known auto-capture/shell noise prefix, or
    (b) a structural notification/tool tag that never appears in a real memory."""
```

A match returns `QualityResult(score=0, quality="low", rejected=True,
rejection_reason="Machine-noise rejected: …")` — same shape as the existing rejection gates.

### Why upstream benefits

The denylist is derived from measured false-accepts, not guesswork, and it targets exactly the content
this project's own hooks generate. Every entry is a structural marker (a tag or a fixed prefix), so the
false-positive surface on genuine prose is small.

## Evidence

| Check | Result |
|---|---|

- **No new tests** (suite unchanged at 6535). This is the weakest of the eight on that axis: it is a
  rejection gate with no test asserting either that the noise is rejected or that ordinary prose is not.
  Recommend requiring a test before merge — happy to add one.
- **Hardcoded list.** The markers are Claude-Code-harness-specific (`<system-reminder>`,
  `task-notification`, `tool-use-id`). Reasonable for this project, but a reviewer may prefer it
  config-driven (`write_gate.noise_patterns`) so other harnesses can extend it without a code change.
- A memory that legitimately *quotes* one of these tags (e.g. a note about debugging the hooks) would be
  rejected. Rare, and the rejection is explicit with the reason in the message, not silent.
- The commit was reformatted with `ruff format` while preparing this branch (quote-style normalisation in
  `_NOISE_TAG_RE`); author preserved.

### DIFF
--- a/src/surreal_memory/engine/quality_scorer.py
+++ b/src/surreal_memory/engine/quality_scorer.py
@@ -67,6 +67,36 @@ _GENERIC_FILLER = frozenset(
     }
 )
 
+# Machine-output / notification noise — never valuable knowledge. Empirical denylist
+# from gate_decision false-accepts (2026-07-23): 180x "Session activity:", 30x <summary>,
+# 20x "Background command", 19x <status>, 12x task-notification, tool-use-id, *-caveat>, etc.
+_NOISE_TAG_RE = re.compile(
+    r"</?summary>|</?status>|</?task-notification>|<system-reminder"
+    r"|[a-z][a-z_-]*-caveat>|<tool_use_id>|tool-use-id>|<function_calls>"
+    r'|\bMonitor event\b|Background command "',
+    re.IGNORECASE,
+)
+_NOISE_PREFIXES = (
+    "session activity:",
+    "<task-notification",
+    "ps -o ",
+    "echo ",
+    "curl -",
+)
+
+
+def _is_machine_noise(content: str) -> bool:
+    """True if content is machine output / session-notification noise, not knowledge.
+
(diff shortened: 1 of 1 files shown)


acidkill:
The denylist is empirically grounded rather than guessed, which is what makes it defensible — 180 false-accepts of `Session activity:` is not a hypothetical.

I hit exactly this while working on the stack today: a recall returned `Session activity: Zrobione. Zweryfikowane...` as if it were knowledge. That is the failure mode this closes.

One follow-up rather than a blocker: `_is_machine_noise` is pure regex logic with no test pinning it. A denylist is exactly the kind of thing that silently rots — a future edit to `_NOISE_TAG_RE` can widen it into real content without anything failing. A couple of cases (one accepted, one rejected per signal) would be cheap insurance. Happy to take it as a separate PR.

Merging.

PR #95, opened 2026-07-26
build(mypy): ignore missing stubs for sentence_transformers

RobertSigmundsson:
### Problem

`sentence_transformers` ships no type stubs. It is absent from the existing
`[[tool.mypy.overrides]]` block that already lists `networkx.*`, `chromadb.*`, `openpyxl.*`,
`google.genai.*` and `watchdog.*` — so in any environment where the package **is** installed, `mypy`
reports import errors for it while every comparable dependency is exempt.

### Change

One line in `pyproject.toml`:

```toml
[[tool.mypy.overrides]]
module = [
    ...
    "watchdog.*",
+   "sentence_transformers.*",
]
ignore_missing_imports = true
```

### Why upstream benefits

Consistency with every other stub-less optional dependency already handled the same way, and a clean
`mypy` run for contributors who have the package installed.

## Evidence

| Check | Result |
|---|---|

- Effectively none. The gate passes with and without it in an environment where the package is not
  installed; the change only matters where it is.
- Worth merging first — it is a one-liner and it is a prerequisite for a clean `mypy` on any branch that
  imports `sentence_transformers`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -258,6 +258,7 @@ module = [
     "openpyxl.*",
     "google.genai.*",
     "watchdog.*",
+    "sentence_transformers.*",
 ]
 ignore_missing_imports = true
 


acidkill:
Correct — the `reranker` extra pulls in sentence-transformers, which ships no stubs, so mypy had no way to pass on a machine with that extra installed. One line, right place.

Merging.

PR #124, opened 2026-07-31
Stop the hook idempotency tests from running against the live database

acidkill:
## Summary

`tests/unit/test_hook_capture_idempotency.py` overrode `HOME` and `SURREAL_MEMORY_BRAIN` in
its fixtures but let `SURREAL_MEMORY_STORAGE` come from the ambient environment. A shell
exporting `SURREAL_MEMORY_STORAGE=surrealdb` — the normal setup for day-to-day dashboard use
— therefore resolved these tests onto the **live shared backend**, with two consequences:

1. **11 of the 12 tests failed** on a configured developer machine, making `make verify`
   unusable there and capable of masking or fabricating failures during a release gate. It
   did exactly that during the v2.18.1 release prep, where the failures had to be
   investigated and ruled out before the real gate could be trusted.
2. **Every run minted a throwaway `idem*` brain on the shared database.** The file's own
   docstring documented this and shipped a cleanup helper to delete the leftovers, rather
   than preventing them.

CI never saw either problem, because the GitHub runner has none of those variables set — so
the suite was green precisely where the fault could not appear.

Isolation is the fixture's responsibility, not the developer's. All three setups (both
fixtures plus the one test that builds its own environment inline) now pin
`SURREAL_MEMORY_STORAGE=sqlite`, matching what the module docstring always claimed the tests
did. `_cleanup_test_brain()` is kept as a belt-and-braces no-op.

Test-only change — no production code touched, so no version bump.

## Test plan

- [x] `uv run pytest tests/unit/test_hook_capture_idempotency.py` **with** the project
      environment exported: 12 passed (was 11 failed / 1 passed)
- [x] Same file with those variables unset (CI-equivalent): 12 passed — unchanged
- [x] `uv run make verify` **with** the project environment exported: exit 0, 7143 passed,
      coverage 71.95% ≥ 67%, security scan clean. This configuration could not pass before.
- [x] `SELECT id, name FROM brain` on the live dev instance after a full suite run: only
      `default` remains — the suite no longer writes to a real database
- [x] Runtime for the file dropped from ~25s to ~2s, since it is no longer making network
      calls to SurrealDB

### DIFF
--- a/tests/unit/test_hook_capture_idempotency.py
+++ b/tests/unit/test_hook_capture_idempotency.py
@@ -10,12 +10,14 @@ HOME) -- never the shared prod brain. Each test uses a unique brain name and
 session id to avoid cross-test collisions in unified_config's process-wide
 _config / _storage_cache singletons.
 
-Caveat: only HOME/SURREAL_MEMORY_BRAIN are overridden, not
-SURREAL_MEMORY_STORAGE. If a developer's shell already has
-SURREAL_MEMORY_STORAGE=surrealdb exported (for day-to-day dashboard use),
-these tests resolve to the live SurrealDB backend instead of SQLite;
-_cleanup_test_brain() below deletes that throwaway "idem*" brain by id in
-that case so the shared DB never accumulates leftovers.
+The fixtures pin SURREAL_MEMORY_STORAGE=sqlite rather than inheriting it.
+Without that pin a developer shell exporting SURREAL_MEMORY_STORAGE=surrealdb
+(the normal setup for day-to-day dashboard use) resolved these tests onto the
+live shared backend, where they both failed -- 11 of 12, while CI stayed green
+because its runner has no such variable -- and minted a throwaway "idem*" brain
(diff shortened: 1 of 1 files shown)


acidkill:
Superseded by the isolation fix already merged in #123.

That version is strictly better than this one and I'm closing in its favour:

- it uses a single **autouse** fixture, so tests that build their own environment inline are covered automatically — this PR had to pin all three call sites by hand and would have missed any future one;
- it also drops `SURREAL_MEMORY_DIR`, a second independent hole I never found: it outranks `Path.home()` in both `UnifiedConfig` and `capture_state._state_path()`, so config and capture state were being read from and written into the developer's real directory rather than `tmp_path`;
- it removes `_cleanup_test_brain()` outright rather than keeping it as a no-op, which is the right call — leaving it in implies these tests can still reach production, and after the pin they cannot.

Same diagnosis, same single root variable (`SURREAL_MEMORY_STORAGE`), better fix. No changes from this branch are worth carrying over.

PR #168, opened 2026-08-15
fix(consolidation): merge must not absorb a pinned fiber

RobertSigmundsson:
## Summary

- `_merge` now skips any pair where either fiber is pinned, so a pinned fiber is no longer absorbed into a merged one.
- Adds a regression test using a pinned fiber that carries neither pattern marker, so it exercises the new guard rather than the existing one.

## Why

`_merge` already refuses to merge fibers carrying `_habit_pattern` or `_reasoning_pattern`, on the grounds that the merged fiber keeps only `merged_from` and would drop the source metadata. The same argument applies to the pin itself, but the marker guard does not reach it: a fiber pinned for any other reason — a trained knowledge base, or an explicit user pin — is still absorbed on the next consolidation pass, and the pin goes with it.

That leaves pinning inconsistent across lifecycle. Decay honours it, dead-neuron pruning honours it, orphan pruning honours it since the guard was hoisted above both branches — and then merge, which is also lifecycle, does not. A fiber surviving every other pass only to be merged away looks like a gap rather than a deliberate policy, and it is silent: the merged fiber is a legitimate object, nothing errors, and the only symptom is that something pinned is no longer there.

The guard is placed alongside the existing marker check and shares its rationale, so the two read as one policy rather than two special cases.

## Test plan

- [x] `pytest tests/ -m "not stress" -n auto` passes locally — 7018 passed, 142 skipped, 1 xfailed (baseline on `main` is 7017; the delta is this PR's test).
- [x] `ruff check src/ tests/` clean.
- [x] `ruff format --check src/ tests/` clean.
- [x] The new test was proven to fail without the fix: with `engine/consolidation.py` reverted to `main`, it fails on `assert "pinned-1" in remaining` — the pinned fiber is gone and only the merged fiber's new id remains.
- [x] The fixture deliberately carries neither `_habit_pattern` nor `_reasoning_pattern`, with an assertion pinning that. An earlier version of this test carried `_reasoning_pattern` and passed with *and* without the change, because the pre-existing marker guard rescued it — worth stating explicitly, since it is an easy trap to fall back into when extending this test later.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/engine/consolidation.py
+++ b/src/surreal_memory/engine/consolidation.py
@@ -938,6 +938,17 @@ class ConsolidationEngine:
             ):
                 continue
 
+            # Never merge a pinned fiber away, for the same reason as the marker
+            # guard above: the merged fiber carries only `merged_from`, dropping
+            # both the pin itself and the source metadata. That guard covers
+            # pattern fibers; it does not cover the other reasons a fiber is
+            # pinned — a trained knowledge base, or an explicit user pin.
+            # Pinning exists precisely to make a fiber survive lifecycle, and
+            # merge is part of lifecycle, so without this a pinned fiber is
+            # silently absorbed on the next consolidation pass.
+            if fiber_list[i].pinned or fiber_list[j].pinned:
+                continue
+
             set_a = fiber_list[i].neuron_ids
             set_b = fiber_list[j].neuron_ids
             intersection = len(set_a & set_b)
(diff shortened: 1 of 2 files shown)


acidkill (approved):
Verified locally on top of this branch:

- Guard placement is right: inside the candidate-pair loop, before the union, alongside the existing `_verbatim` and `_habit_pattern`/`_reasoning_pattern` guards — a pinned fiber can never enter a merge group.
- The lifecycle argument checks out: `_prune` already skips `pinned_neuron_ids` (hoisted above both the orphan and dead paths) and decay preloads the same set; merge was the one lifecycle path without the guard.
- The claim is proven, not just trusted: with only the two-line guard reverted, the new test fails on `merge must never delete a pinned fiber`; with it, the full consolidation module passes (21/21).
- `Fiber.pinned` is a first-class field (`core/fiber.py:66`), so `fiber_list[i].pinned or fiber_list[j].pinned` covers both members of the pair.
- ruff check / format clean; mypy shows only the pre-existing `google.genai\] env error; Integration rerun is green (the earlier fail was the pre-#172 signin flake, not this diff).

Nice touch on the fixture deliberately avoiding the pattern markers — the test pins the new guard, not the old one.

PR #177, opened 2026-08-15
fix(reasoning): don't apply the per-session injection marker to per-subagent events

RobertSigmundsson:
## Summary

- The per-session injection marker is no longer applied to events that fire once per *subagent* rather than once per session.
- Such events also no longer *set* the marker, so they cannot starve `SessionStart` / `UserPromptSubmit` in turn.
- Adds a regression test pinning both halves.

## Why

`get_reasoning_context()` is the shared entry point behind the reasoning-strategy hooks, and it applies the session-scoped idempotency marker to every caller alike. That is right for `SessionStart` and `UserPromptSubmit`: both fire for the same session and race, so whichever arrives first should win and the other should be a no-op.

It is wrong for any event that fires once per spawned subagent. The marker is keyed on `session_id`, and a subagent payload carries the **parent** session's id — so once the parent has been injected, `already_injected(session_id)` is true for every subagent that parent will ever spawn, and each one gets `""` back.

The failure is silent by construction. `""` is also the legitimate "injection is disabled, or nothing matched" answer, so there is no log line and no error to notice: a long-running session simply stops handing strategies to its subagents after the first one, and nothing says so.

A note on scope, since it affects how you'll want to read this. The repo does not ship a `SubagentStart` hook — there is no such entry point in `pyproject.toml` and no reference to it under `src/`. `get_reasoning_context()` is nonetheless event-agnostic and users wire it to the events they want; `SubagentStart` is simply the case that exposes the mismatch between a session-keyed marker and a per-subagent event. So for a stock install this change is a **no-op**: `hook_event_name` never takes that value through the shipped wiring. I'm sending it because the guard belongs with the marker rather than with whatever wires it, and because the next person to point a per-subagent event at this function would hit the same silent floor. If you'd rather this waited for the repo to ship such a hook, that's a completely reasonable call and I'll withdraw it.

## Test plan

- [x] `pytest tests/ -m "not stress" -n auto` passes locally — 7062 passed, 146 skipped, 1 xfailed (baseline on `main` is 7061; the delta is this PR's test).
- [x] `ruff check src/ tests/` clean.
- [x] `ruff format --check src/ tests/` clean — 730 files already formatted.
- [x] The new test was proven to fail without the fix: with `src/surreal_memory/engine/reasoning_injection.py` reverted to `main` and the test in place, it fails on `assert "learned from claude-fable-5" in block` with `block == ''` — the empty-block starvation this PR removes.
- [x] Diff touches `engine/` and `tests/` only; no server routes or dashboard assets, so no live-instance check applies.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/engine/reasoning_injection.py
+++ b/src/surreal_memory/engine/reasoning_injection.py
@@ -235,19 +235,29 @@ async def get_reasoning_context(hook_input: dict[str, Any]) -> str:
     """Resolve the active model, build its reasoning block, mark the session.
 
     Shared by the SessionStart and UserPromptSubmit hooks. Opt-in via
-    reasoning_training.injection_enabled; injects at most once per session via the
-    marker (already_injected/mark_injected), so whichever hook fires first wins and
-    the other is a no-op. Storage is opened on the current brain and always closed.
-    Returns "" when injection is disabled, already done this session, or nothing
-    matched.
+    reasoning_training.injection_enabled. Those two fire for the same session and
+    race each other, so they inject at most once per session via the marker
+    (already_injected/mark_injected) — whichever gets there first wins.
+
+    The marker is keyed on ``session_id``, so it is only correct for events that
+    fire once per session. A per-subagent event (``SubagentStart``, which a user
+    can wire to this same entry point) carries the PARENT session's id, so
+    honouring the marker there would return "" for every subagent after the
+    first — and silently, since "" is also the legitimate "nothing matched"
+    answer. Such events are exempt from the marker and do not set it.
+
+    Storage is opened on the current brain and always closed. Returns "" when
+    injection is disabled, already done this session (session-scoped events
+    only), or nothing matched.
     """
     from surreal_memory.unified_config import get_config, get_shared_storage
 
     config = get_config()
     if not config.reasoning_training.injection_enabled:
         return ""
+    per_session = str(hook_input.get("hook_event_name") or "") != "SubagentStart"
     session_id = str(hook_input.get("session_id") or "")
-    if already_injected(session_id):
+    if per_session and already_injected(session_id):
         return ""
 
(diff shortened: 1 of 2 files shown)


acidkill (approved):
Thanks for this — clean root-cause, clean fix, and thanks for calling out the scope explicitly in the description (stock installs get a no-op, confirmed: `SubagentStart` isn't wired to any shipped hook entry point, so this is purely forward-looking for anyone who wires it themselves).

I ran this through six independent review passes (correctness, comments, test coverage, silent-failure risk, type design, simplification) plus my own read of the diff. Summary:

**Correctness** — the `per_session` gate correctly closes both halves: it bypasses the read-check (`already_injected`) *and* skips the write (`mark_injected`) for `SubagentStart`, so a subagent event can't starve itself either. Traced `session_id` and `hook_event_name` through the rest of the file — nothing else touches either in a way that would still misbehave per-subagent, and `resolve_active_model` doesn't read either field, so there's no interaction risk with model resolution.

**Docstring** — matches the new code exactly, including *why* the marker is wrong for a parent-scoped id.

**Tests** — the new test proves both halves for real, not trivially: the parent's marker is genuinely pre-set on disk before the bypass-on-read assertion, and the no-write assertion uses a fresh, previously-unmarked session id, so a regression that dropped the `per_session` guard on the write side would actually fail it. Good reuse of the existing `_ucfg`/`_add_pattern`/`InMemoryStorage` fixtures.

**One non-blocking observation**: the guard is an exact string match against `"SubagentStart"` with no logging on a mismatch, so a future typo'd or renamed event on Claude Code's side would silently fall back to the pre-fix behavior for that event. That's a scope boundary, not a regression — every event that isn't recognized today already has that behavior — just flagging it as something to keep in mind if/when a second event needs the same exemption.

CI is green across the board. Merging — thanks again for the contribution, @RobertSigmundsson.

PR #206, opened 2026-09-05
fix(storage): brain import surfaces per-loop drops instead of swallowing them

RobertSigmundsson:
## Summary

- `SurrealDBStorage.import_brain` now counts what each of its three per-record loops actually wrote, and warns with both numbers when the totals disagree, instead of discarding failures silently.
- `smem brain import` reports the destination brain's own counts rather than the size of the file it was given.
- Adds a regression test covering both the shortfall warning and the silence of a clean import.

## Why

`import_brain` wraps each of its three per-record loops — neurons, synapses, fibers — in a bare `except Exception: pass`:

```python
# src/surreal_memory/storage/surrealdb/store.py, import_brain()
for nd in snapshot.neurons:
    try:
        neuron = Neuron(...)
        await self.add_neuron(neuron)
    except Exception:
        pass
```

Anything failing to write during an import therefore vanished: no counter, no log line, nothing in the return value. A duplicate record id is the trivial way in, since `_to_surreal_id` folds a neuron's own id into `neuron:<id>` with no brain component — so importing a snapshot under a fresh `--name` whilst the original brain still holds those ids raises straight into the except.

The two later loops in the *same function*, added for projects and typed memories by #134, already handle this properly with `logger.debug("Skipped unimportable project record", exc_info=True)`. The pattern simply never propagated up to the first three loops.

`smem brain import` compounded it by never asking the storage layer what landed:

```python
# src/surreal_memory/cli/commands/brain.py, brain_import()
await storage.import_brain(snapshot, storage.brain_id)   # return value discarded
...
typer.echo(f"  Neurons: {len(data['neurons'])}")          # the source file, not the destination
```

Exit code 0 and a confident count, even if the destination received nothing. The HTTP route already gets this right — `server/routes/brain.py` calls `storage.get_stats(imported_id)` and reports real numbers — so the CLI was the odd one out on a report the operator is more likely to be reading.

## Changes

- `storage/surrealdb/store.py::import_brain`: per-loop success counters; `logger.debug(..., exc_info=True)` on each swallow, so the exception is no longer thrown away either; and `logger.warning("Imported %d of %d neurons into brain %r; %d dropped", …)` after a loop whose shortfall is non-zero. The warning carries both counts and the target brain id, so the log can be read without the input file to hand.
- `cli/commands/brain.py::brain_import`: captures the returned brain id, calls `storage.get_stats(imported_bid)` in a `try` with `close()` in the `finally`, and prints `neuron_count`, `synapse_count` and `fiber_count` in place of `len(data[...])`.
- `tests/unit/test_surrealdb_import_brain_reports_drops.py` (new, 140 lines): `test_import_brain_warns_when_neurons_are_dropped` swaps `add_neuron` for a callable that raises on its second call and asserts a `WARNING` naming both counts, plus that `import_brain` still returns the target brain id; `test_import_brain_silent_when_everything_lands` asserts no warning when every write succeeds.

## A deliberate choice worth naming

`get_stats` sits in the `try`, not in a `except: pass` of its own. If the verification query fails, `close()` still runs and the command reports the failure rather than falling back to the file-derived numbers. Printing counts that might be wrong is the behaviour this PR exists to remove, so it seemed the wrong place to be forgiving. Say the word if you would rather it degraded to a warning and the old output.

The return type stays `str`. `NeuralStorage.import_brain -> str` in `storage/base.py`, the HTTP route, and `InMemoryBrainMixin.import_brain` (which never had this bug — it has no per-record `try/except` at all) are all untouched. This is a report-loudly change, not an interface change.

## Not touched, on purpose

- `_to_surreal_id`. The `neuron:<id>` folding is *why* collisions are easy, but re-scoping ids by brain reaches every read and write path, the snapshot format and the dashboard queries, and would invalidate every existing brain on disk. That is a separate conversation; this PR only stops the drops being silent.
- Raising on a shortfall. The storage layer now warns and the CLI now prints the true counts; whether a partial import should be a hard failure is a policy call for the caller.

## Test plan

- [x] `pytest tests/unit/test_surrealdb_import_brain_reports_drops.py` — 2 passed.
- [x] With both source files reverted to `main` and the new test kept, it fails on `AssertionError: expected 'imported 2 of 3 neurons' warning; got: []`, whilst `test_import_brain_silent_when_everything_lands` still passes — so the pair measures this change rather than itself.
- [x] `pytest tests/ -m "not stress" -n 4` against a live SurrealDB v3.2.0 — 7285 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus exactly the two tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.

The CLI change itself has no test. The report is three `typer.echo` lines over a dict the storage layer already covers, and the existing `brain import` tests do not drive a live backend; adding a harness for it looked disproportionate to three print statements. Happy to add one if you would rather it were pinned.

## Related issues

None filed; this is the fix rather than the report. The nearest neighbour is issue #174, which described the identical *shape* — a bare `except: pass` swallowing a write — in `add_neuron`'s own `neuron_state` insert. #134 added the logging loops for projects and typed memories in this same function. Neither covers these three loops. I searched issues and pull requests for `import_brain`, `brain import`, `silent`, `collision`, `duplicate id` and `swallow` without finding anything else on it.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/cli/commands/brain.py
+++ b/src/surreal_memory/cli/commands/brain.py
@@ -345,17 +345,26 @@ def brain_import(
 
         # Load/create storage and import
         storage = await get_storage(config, brain_name=brain_name)
-        await storage.import_brain(snapshot, storage.brain_id)
-        await storage.close()
+        imported_bid = await storage.import_brain(snapshot, storage.brain_id)
+        # Report the counts the storage actually ended up with, not the counts
+        # the source file said were in the payload. `_to_surreal_id` folds
+        # neuron ids to `neuron:<id>` without a brain component, so importing
+        # under a fresh name can still collide with an existing brain's ids;
+        # `import_brain` now logs the shortfall, and this report tells the
+        # operator what actually landed instead of what the file claimed.
+        try:
+            stats = await storage.get_stats(imported_bid)
+        finally:
+            await storage.close()
 
         if use:
             config.current_brain = brain_name
             config.save()
 
         typer.secho(f"Imported brain: {brain_name}", fg=typer.colors.GREEN)
-        typer.echo(f"  Neurons: {len(data['neurons'])}")
-        typer.echo(f"  Synapses: {len(data['synapses'])}")
-        typer.echo(f"  Fibers: {len(data['fibers'])}")
+        typer.echo(f"  Neurons: {stats['neuron_count']}")
+        typer.echo(f"  Synapses: {stats['synapse_count']}")
+        typer.echo(f"  Fibers: {stats['fiber_count']}")
 
     run_async(_import())
 
--- a/src/surreal_memory/storage/surrealdb/store.py
+++ b/src/surreal_memory/storage/surrealdb/store.py
@@ -2529,6 +2529,14 @@ class SurrealDBStorage(
         )
         await self.save_brain(brain)
 
+        # Count successful writes per loop. Bare `except: pass` used to swallow
+        # duplicate-id collisions (trivially reachable — `_to_surreal_id` folds
+        # neuron ids to `neuron:<id>` without a brain component, so importing a
+        # snapshot under a fresh name while the original brain still holds those
+        # ids raises straight into these except blocks). The shortfall now
+        # surfaces as a warning; the return type stays `str` so the abstract
+        # method contract in NeuralStorage does not change.
+        neuron_ok = 0
         for nd in snapshot.neurons:
             try:
                 neuron = Neuron(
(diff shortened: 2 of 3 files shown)


acidkill (approved):
Reviewed the full diff. Per-loop drop counters with a once-per-loop warning, and the CLI now reports the counts the database actually ended up with instead of the snapshot's claims — the right fix given _to_surreal_id's brain-less neuron ids.

PR #207, opened 2026-09-05
fix(cli): listing previews show essence for GRAPH_ONLY anchors; tombstone literal becomes a shared constant

RobertSigmundsson:
## Summary

- `smem list` and `smem cleanup --dry-run` no longer render the `GRAPH_ONLY` tombstone as a memory's preview: the four preview call sites fall through to `fiber.essence` when the anchor holds the placeholder.
- The placeholder literal becomes `core.constants.GRAPH_ONLY_PLACEHOLDER`, imported everywhere it is compared — follow-up 1 from the review on #193.
- Adds nine tests, two of which drive `list_memories` itself rather than the helper in isolation.

## Why

The four call sites in `cli/commands/listing.py` walked the preview fallback only as far as the anchor neuron:

```python
# src/surreal_memory/cli/commands/listing.py — the same shape in all four sites
if fiber.summary:
    content = fiber.summary
elif fiber.anchor_neuron_id:
    anchor = await storage.get_neuron(fiber.anchor_neuron_id)
    if anchor:
        content = anchor.content
# fiber.essence is never consulted
```

That misses precisely the case compression creates. At the `GRAPH_ONLY` tier, `CompressionEngine.compress_fiber` overwrites the anchor's content with the sentinel through its `pending_clear` batch. The write does not touch `fiber.essence`, and `ConsolidationEngine._essence_backfill` may already have produced a real essence from the anchor's original text before compression ran. A fiber with no summary, a tombstoned anchor and a perfectly serviceable essence therefore listed as `[graph-only]` — indistinguishable, to someone skimming, from genuinely cryptic content. `cleanup --dry-run` printed the same unhelpful preview immediately before offering to delete the thing.

## The constant, and the count

Follow-up 1 on the #193 review asked for this directly: the literal had, in the reviewer's count, reached *"fourteen occurrences across eight"* files, every guard comparing against a bare string, so changing the placeholder text in `compression.py` *"would type-check, lint clean, and silently disable every guard"*. All fourteen now import `GRAPH_ONLY_PLACEHOLDER` from `core/constants.py`. The fix above needed a fifteenth comparison in `listing.py`; it imports the constant rather than becoming a ninth file with a literal in it.

The test suite still compares against the raw string in several places. That is deliberate — those are the assertions that ought to fail loudly if the value ever changes, which is the whole point of the exercise.

## Changes

- `core/constants.py` (new): `GRAPH_ONLY_PLACEHOLDER`, with a docstring pointing at the write site.
- `engine/`: `compression.py`, `consolidation.py`, `dedup/pipeline.py`, `interference.py`, `pipeline_steps.py`, `prediction_error.py`, `retrieval.py`, `semantic_discovery.py` — every literal replaced, one import per file.
- `cli/commands/listing.py`: new `_fiber_preview_content(fiber, storage)` helper ordering `fiber.summary` → `anchor.content` (skipping the placeholder) → `fiber.essence` → `""`, used by all four sites: `list_memories`'s expired, fiber-fallback and typed-memory branches, and `cleanup --dry-run`'s preview. The cleanup site still truncates to `[:50]`.
- `tests/unit/test_listing_fallback_essence.py` (new): seven helper-level tests covering each branch and guard, plus `test_expired_branch_renders_essence_not_tombstone` and `test_typed_branch_renders_essence_not_tombstone`, which go through `list_memories` with patched storage and assert on the rendered output.

The helper keeps `summary → anchor → essence` rather than adopting `adapters/langchain.py`'s `anchor → summary → essence`, which is tuned for retrieval. Unifying the two would change what `smem list` shows for every fiber that has both a summary and anchor content — a separate decision from this bug. The helper's docstring records the divergence.

## Not touched, on purpose

- Compression's write of the placeholder. The tombstone is deliberate; only the rendering was wrong.
- `typed_memory.content`, defined in `storage/surrealdb/schema.py` but never written and never read. It came up alongside this bug, but it is a schema question on a `SCHEMALESS` table and unrelated to the preview chain. Happy to raise it separately if it is worth anyone's time.
- `adapters/langchain.py` and `engine/fidelity.py`, which already consult all three fields in orders suited to what they render.

## Test plan

- [x] `pytest tests/unit/test_listing_fallback_essence.py` — 9 passed.
- [x] Differential control, measured the strict way: keeping the helper and reverting **only the four call sites** to the old two-step chain leaves 7 passed and fails both call-site tests, which render `[graph-only]`. An earlier version of this test file passed 7/7 in that state, which is why the two call-site tests exist.
- [x] `pytest tests/unit/test_listing_fallback_essence.py tests/unit/test_compression.py tests/unit/test_consolidation.py` — 119 passed.
- [x] `pytest tests/ -m "not stress" -n 4` against a live SurrealDB v3.2.0 — 7292 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus exactly the nine tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] `mypy src/ --ignore-missing-imports` — success, no issues found in 355 source files, one more than `main` for the new module.
- [x] Coverage under the CI gate: 72.64%, against 72.36% on `main`.

## Related issues

Follow-up 1 from the #193 review is done here. Follow-up 2 from the same review — the unlogged `except` around the brain pre-fetch in `compression.py` — is a separate one-line change and gets its own PR rather than riding along in this diff.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/cli/commands/listing.py
+++ b/src/surreal_memory/cli/commands/listing.py
@@ -8,10 +8,42 @@ from typing import Annotated, Any
 import typer
 
 from surreal_memory.cli._helpers import get_config, get_storage, output_result, run_async
+from surreal_memory.core.constants import GRAPH_ONLY_PLACEHOLDER
 from surreal_memory.core.memory_types import MemoryType, Priority
 from surreal_memory.safety.freshness import evaluate_freshness, format_age
 
 logger = logging.getLogger(__name__)
+
+
+async def _fiber_preview_content(fiber: Any, storage: Any) -> str:
+    """Preview text for a fiber: ``summary → anchor.content → essence``.
+
+    Behaviour-preserving over the previous two-step chain
+    (``summary → anchor.content``) for every case except one: when the anchor
+    has been reduced to the ``GRAPH_ONLY`` compression sentinel
+    (``"[graph-only]"``) the helper falls through to ``fiber.essence``.
+    ``_essence_backfill`` in ``engine/consolidation.py`` may have generated a
+    real essence from the anchor's original content before compression ran,
+    so the tombstone is precisely the case where essence is the correct thing
+    to render.
+
+    Kept in this module (rather than folded into ``adapters/langchain.py``'s
+    ``_page_content``) because langchain's helper uses a different order
+    (``anchor → summary → essence``) tuned for retrieval, and unifying the
+    two would change ``smem list``'s user-visible behaviour beyond the
+    scope of this fix.
+    """
+    if fiber.summary:
+        return str(fiber.summary)
+    if fiber.anchor_neuron_id:
+        anchor = await storage.get_neuron(fiber.anchor_neuron_id)
+        if anchor is not None:
+            content = getattr(anchor, "content", None)
+            if content and content != GRAPH_ONLY_PLACEHOLDER:
+                return str(content)
+    if fiber.essence:
+        return str(fiber.essence)
+    return ""
 
 
 def list_memories(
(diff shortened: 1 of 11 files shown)


acidkill (approved):
Reviewed the full diff. Shared GRAPH_ONLY_PLACEHOLDER constant replaces fourteen scattered literals, and the listing preview gains the essence rung exactly where the tombstone used to surface. Call-site tests cover both listing branches, not just the helper. Green CI.

PR #208, opened 2026-09-05
fix(storage): address record rows by id, not by a rebuilt string

RobertSigmundsson:
## Summary

- Seven single-row lookups across `alerts.py`, `cognitive.py` and `versions.py` compared a record id with a Python string, a predicate that is unconditionally false, so none of them could ever match a row.
- The writes beside them now reuse the id their own SELECT returned instead of rebuilding it, which silently addressed a different record for letter-free ids.
- Row mappers now hand back an id the caller can pass straight back in, rather than one carrying SurrealDB's quoting.
- Adds eighteen live tests, each seeded with a decoy row.

## Why

The pattern was the same in all seven: bind `rid = f"<table>:{sid}"` and compare it with `id = $rid`. `id` holds a record id, not a string, so the comparison is false regardless of the data. These were not intermittent failures but structurally dead paths:

- `alerts.py` — an alert could never leave `active`, acknowledging one reported "not found", and reading a single alert always returned `None`.
- `cognitive.py` — a knowledge gap could not be read back or closed, both of which `mcp/cognitive_handler.py` calls.
- `versions.py` — `get_version` returned `None`, so every restore and diff path in `engine/brain_versioning.py` saw an empty result.

They now bind the sanitised id and rebuild the record id in SurrealQL with `type::record('<table>', $sid)`.

## The second half: rebuilt ids write into the void

Fixing the lookups alone would have left a worse bug in place. The writes rebuilt `f"<table>:{sid}"` too, and for an all-digit sid that parses back through the SDK as a *numeric* record id whilst the row was stored under a *string* one. The merge then writes to a different, empty record and reports success having changed nothing — which is harder to notice than the failure it replaces.

Measured on SurrealDB 3.2.0, against a row created with a string key:

```
UPDATE probe:1234567890123456 MERGE { v: 'after' };   -- no error
SELECT v FROM probe;                                  -- still 'before'
```

The same shape applied to `update_cognitive_evidence` and `delete_version`; `save_version`'s insert-retry had all three symptoms at once — a delete that cleared nothing, an `except: pass` that said nothing, and a retry straight back into the same collision.

## The third half: the id handed to the caller

SurrealDB renders a record id whose id part carries no letter in its quoted form — `alerts:⟨1122334455667788⟩`. The row mappers stripped only the table prefix, so the guillemets reached the caller, and `_to_surreal_id` then mapped them to underscores and produced an id matching nothing. `get_active_alerts` listed an alert that `mark_alert_acknowledged` then called missing.

Measured on 3.2.0: the quoted form appears for `0001`, `1122334455667788`, `12_34` and `_123`, and not for `a1`, `123abc`, `1_2a` or `_a1` — exactly when the id carries no letter. An `uuid4().hex[:16]` sid is letter-free about once in 1150: character 12 is always the version nibble `4`, so fifteen of the sixteen characters are random and the chance is `(10/16)**15`. A four-million-draw check gives one in 1130. `str(uuid4())` ids essentially never qualify, but they pass through the same mappers.

`_record_id_part` in `_ids.py` is the one place that knows the inverse of the sanitiser, and it sits beside it. It is deliberately not called `_from_surreal_id`: `store.py` already has a function by that name which also maps `_` back to `-`, which is a different question.

## Test plan

- [x] `pytest tests/unit/test_surrealdb_alerts_recordid_live.py tests/unit/test_surrealdb_record_id_lookups_live.py` — 18 passed against a live SurrealDB v3.2.0.
- [x] Differential control, stated precisely rather than roundly: with the three source files reverted to `main` and the tests kept, **14 of the 18 fail**. The four that pass are the controls and could not have failed — `test_unknown_ids_are_still_reported_as_missing`, `test_unknown_gap_id_is_reported_as_missing` and `test_unknown_version_id_is_reported_as_missing` assert that an unknown id is still reported missing, and `test_resolve_alerts_by_type_still_works` is the positive control for the one alert path that was never broken.
- [x] Each test seeds a decoy row alongside its target. Against a single-row table an unfiltered lookup is indistinguishable from a correct one, so without the decoy these tests would pass just as happily for a build with the id filter deleted.
- [x] `pytest tests/ -m "not stress" -n 4` — 7301 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus exactly the eighteen tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] Coverage under the CI gate: 72.70%, against 72.36% on `main`.

## A note for whoever merges second

`tests/unit/_surrealdb_live.py` conflicts textually with the `change_log` FLEXIBLE fix and the sync-reader change that is stacked on it: both append a brain name to `LIVE_TEST_BRAIN_NAMES` immediately after the same entry. The resolution is to keep both lines; I will rebase whichever of us lands second.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/storage/surrealdb/_ids.py
+++ b/src/surreal_memory/storage/surrealdb/_ids.py
@@ -45,6 +45,28 @@ def _to_surreal_id(record_id: str) -> str:
     )
 
 
+def _record_id_part(record_id: str) -> str:
+    """Bare id part of a record id, as ``_to_surreal_id`` would have produced it.
+
+    The inverse direction of the choke point above, and the reason it belongs
+    here rather than being spelled out per mixin: SurrealDB renders a record id
+    whose id part contains **no letter** in its quoted form —
+    ``alerts:⟨1122334455667788⟩`` — and a plain ``split(":")[-1]`` hands those
+    guillemets back to the caller. Feeding that back into ``_to_surreal_id``
+    maps them to underscores, so the id no longer addresses its own row: the
+    lookup misses and the caller is told the record does not exist.
+
+    Measured on SurrealDB 3.2.0: the quoted form appears for ``0001``,
+    ``1122334455667788``, ``12_34`` and ``_123``, but not for ``a1``,
+    ``123abc``, ``1_2a`` or ``_a1`` — i.e. exactly when the id carries no
+    letter. Ids minted as ``str(uuid4())`` almost always carry one; ids minted
+    as ``uuid4().hex[:16]`` (alerts) are letter-free about once in 1150.
+    """
+    if ":" in record_id:
+        record_id = record_id.rsplit(":", 1)[1]
+    return record_id.strip("⟨⟩`")
+
+
 def _safe_brain_id(brain_id: str) -> str:
     """Validate a brain id before it is inlined raw into a record id / SurQL.
 
--- a/src/surreal_memory/storage/surrealdb/alerts.py
+++ b/src/surreal_memory/storage/surrealdb/alerts.py
@@ -1,4 +1,25 @@
-"""SurrealDB alerts storage mixin."""
+"""SurrealDB alerts storage mixin.
+
+Single-alert lookups bind the *sanitised* id and rebuild the record id inside
+SurrealQL with ``type::record('alerts', $sid)``. Comparing ``id`` against a
+plain ``"alerts:<sid>"`` string is not a slower spelling of the same thing —
+``id`` holds a record id, so the predicate is unconditionally false and the row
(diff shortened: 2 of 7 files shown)


acidkill (approved):
Reviewed the full diff. Rebuilding record ids in SurrealQL via type::record() and reusing the id object the SELECT returned — the string-comparison predicates were structurally dead and the numeric-id rebuild could address a different row. The _record_id_part guillemet handling matches the measured SDK behaviour. Live tests cover the dead lookups plus a positive control.

PR #209, opened 2026-09-05
fix(consolidation): the lock honours SURREAL_MEMORY_DIR like the rest of the data dir

RobertSigmundsson:
## Summary

- `_lock_path` now honours `SURREAL_MEMORY_DIR`, so the consolidation lock moves with the data directory like everything else in the package.
- The lock test class gets a per-test directory and a test that asserts the redirection rather than assuming it.

## Why

`_lock_path` derived its directory from `Path.home()` alone. Everything else already reads the variable — `cli/config.get_default_data_dir` and `cli/update_check._get_cache_path` both do — which made the consolidation lock the one piece of state a redirected data dir could not move. A run against a throwaway brain still created and deleted files in the operator's real `~/.surrealmemory`.

Measured by asking `_lock_path` where it would put the file, with the variable set:

```
main:   SURREAL_MEMORY_DIR=/tmp/throwaway  ->  ~/.surrealmemory/consolidation-<brain>.lock
here:   SURREAL_MEMORY_DIR=/tmp/throwaway  ->  /tmp/throwaway/consolidation-<brain>.lock
```

## A trade-off, named rather than discovered

Two processes pointed at different `SURREAL_MEMORY_DIR` values now take different locks and no longer exclude one another. That is the intended reading — they are operating on separate data directories — but it does mean the lock is per data directory rather than per machine. If you would rather it stayed machine-wide, this is the wrong shape and I would rather know now.

## Changes

- `utils/consolidation_lock.py`: read `SURREAL_MEMORY_DIR`, falling back to `Path.home()/".surrealmemory"` as before, with the trade-off recorded in the docstring.
- `tests/unit/test_multi_agent.py`: the autouse fixture points `SURREAL_MEMORY_DIR` at a per-test directory, and `test_lock_lives_in_the_configured_data_dir` asserts the redirection. The failure mode is silent — the lock simply reappears in the real home — so it is worth an explicit assertion.

## A correction to that class's own docstring

The docstring says these tests race on one shared lock file under `xdist`, and that `xdist_group` pins the class to a single worker to prevent it. That was true when #74 wrote it, and is no longer true of the suite as it stands — which is worth stating carefully, because the note was not wrong, it was overtaken.

#121 later added a session-scoped `$HOME` redirect to `conftest` for unrelated reasons. That gives each worker its own home directory, and therefore its own lock file, so the race stopped being reachable and nothing updated the note. Checked directly: with `$HOME` redirected the lock lands under the fake home on `main` too, before any of this change.

What the redirect does not do is make the intent explicit. A test relying on `$HOME` being someone else's problem is one refactor away from writing to a real home again, which is why the fixture now says so in its own terms. `xdist_group` is kept as a guard against shared state being reintroduced, not as a live fix.

## Test plan

- [x] `pytest tests/unit/test_multi_agent.py` — 16 passed.
- [x] With `utils/consolidation_lock.py` reverted to `main` and the tests kept, exactly one fails: `test_lock_lives_in_the_configured_data_dir`. The other fifteen pass either way, which is the point — this change is meant to move a file, not alter locking behaviour.
- [x] `pytest tests/ -m "not stress" -n 4` — 7284 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the single test added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.

## Verified by

@RobertSigmundsson

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -385,7 +385,7 @@ places, the environment wins.
 | `SURREAL_MEMORY_API_KEY` | `src/surreal_memory/unified_config.py` |
 | `SURREAL_MEMORY_BRAIN` | `src/surreal_memory/cli/_helpers.py`, `src/surreal_memory/cli/commands/brain.py`, `src/surreal_memory/unified_config.py` |
 | `SURREAL_MEMORY_DASHBOARD_CACHE_TTL` | `src/surreal_memory/server/dashboard_cache.py` |
-| `SURREAL_MEMORY_DIR` | `src/surreal_memory/cli/config.py`, `src/surreal_memory/cli/update_check.py`, `src/surreal_memory/engine/reasoning_injection.py`, +3 more |
+| `SURREAL_MEMORY_DIR` | `src/surreal_memory/cli/config.py`, `src/surreal_memory/cli/update_check.py`, `src/surreal_memory/engine/reasoning_injection.py`, +4 more |
 | `SURREAL_MEMORY_DISABLE_SUPERSEDED_FILTER` | `src/surreal_memory/mcp/recall_handler.py` |
 | `SURREAL_MEMORY_EMBEDDING_API_KEY` | `src/surreal_memory/engine/embedding/bge_m3_embedding.py` |
 | `SURREAL_MEMORY_EMBEDDING_DIMENSION` | `src/surreal_memory/engine/embedding/bge_m3_embedding.py`, `src/surreal_memory/unified_config.py` |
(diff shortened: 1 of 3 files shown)


acidkill (approved):
Reviewed the full diff. The consolidation lock now honours SURREAL_MEMORY_DIR like every other piece of data-dir state, and the tests assert the redirection instead of assuming it.

PR #210, opened 2026-09-05
fix(consolidation): LIFECYCLE reads access_frequency from NeuronState, not metadata

RobertSigmundsson:
## Summary

- The LIFECYCLE heat-score pass now reads `access_frequency` and `last_activated` from `NeuronState`, where they live, instead of from `neuron.metadata`, where they never did.
- Adds two tests: one that the value reaches `calculate_heat_score`, and one covering the fallback path that had no coverage at all.

## Why

`_lifecycle` read both fields off `neuron.metadata`. Neither is stored there — both are columns on `neuron_state`, keyed by `neuron_id` — and `_row_to_neuron` copies only the JSON `metadata` blob, so the `neuron` table's own columns of those names never reach the dataclass either.

The reads therefore returned their defaults for every neuron on every pass. `calculate_heat_score` weights access at 0.4 and recency at 0.4 against priority's 0.2, so eighty per cent of the calculation was pinned to zero and heat collapsed to `priority * 0.2`. A neuron recalled a thousand times an hour aged into COOL, COMPRESSED and ARCHIVED on exactly the same schedule as one nobody had ever touched — the tiering worked, it was simply being fed constants.

The fix borrows the pattern the dead-neuron and orphan pass in the same class already uses: a brain-wide prefetch through `get_all_neuron_states`, with a `get_neuron_states_batch` fallback when that call is unavailable. `priority` stays on `neuron.metadata`, which is where the writer puts it.

## A behaviour change worth flagging

On an existing brain the first LIFECYCLE pass after this lands will recompute `lifecycle_state` for everything, because eighty per cent of the heat weight comes back to life at once. Neurons that were quietly demoted on a constant will be re-evaluated against what `NeuronState` actually records, and `compression_tier` follows `lifecycle_state`. That is the intended correction rather than a side effect, but it is a one-off reshuffle rather than a quiet improvement, and it seemed better said than discovered.

One caveat on what "actually records" means, since it changes how the result should be read: `last_activated` is also stamped by the decay pass, and a comment in that pass already notes that a decay pass is not usage. So `recency_score` will read more recent than pure recall history would suggest, and with the default weights a neuron with a few recalls behind it will tend to stay warm. That is a pre-existing property of the field rather than something this change introduces, but this change is what makes it visible.

## About the fallback, and what it costs

The fallback is one query for the whole pass — not one per page, and certainly not one per neuron. `neurons` is already the entire accumulated set by the time it runs, so the cache is filled on the first miss and never touched again. On a brain at the 10 000 cap that single call is the fallback's whole cost, which is worth knowing: the prune pass in the same file measures roughly ten seconds per five thousand ids against the same backend, and this runs inside a per-strategy timeout. The comment in the code now says so.

An earlier draft of this change described that fallback as "per-page", and the variables were named accordingly. They were not paging anything. Both are corrected here rather than left to mislead the next reader.

## Test plan

- [x] `pytest tests/unit/test_lifecycle_heat_score.py` — 2 passed.
- [x] `test_lifecycle_reads_access_frequency_from_neuron_state` spies on `calculate_heat_score` and asserts `access_count` is 42 for a neuron whose `NeuronState.access_frequency` is 42, and 0 for the control. Without the fix the spy sees zero for both.
- [x] `test_lifecycle_falls_back_to_one_batch_when_the_prefetch_fails` makes `get_all_neuron_states` raise and asserts `get_neuron_states_batch` is called exactly once, covers every id being scored, and yields the same answer the prefetch would have. That branch was previously reached by no test.
- [x] With `engine/consolidation.py` reverted to `main` and both tests kept, both fail.
- [x] `pytest tests/ -m "not stress" -n 4` — 7285 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the two tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] Coverage under the CI gate: 72.42%, against 72.36% on `main`.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/engine/consolidation.py
+++ b/src/surreal_memory/engine/consolidation.py
@@ -2803,18 +2803,49 @@ class ConsolidationEngine:
         config = CompressionConfig()
         states_updated = 0
 
-        for neuron in neurons:
-            # Retrieve last_accessed_at and access_frequency from neuron metadata
-            # (access_frequency is stored in neuron_states, not neurons directly)
-            last_accessed_raw: str | None = neuron.metadata.get("last_accessed_at")
-            last_accessed_at: datetime | None = None
-            if last_accessed_raw:
-                try:
-                    last_accessed_at = datetime.fromisoformat(last_accessed_raw)
-                except ValueError:
-                    pass
+        # access_frequency and last_activated live on NeuronState (schema.py neuron_state
+        # table), not on neuron.metadata — the dead-neuron / orphan pass in this same
+        # class already knows this (_prune, ~L1000) and prefetches with the exact
+        # pattern below. Without this, access_score and recency_score (0.8 of the heat
+        # weight) were pinned to zero and heat reduced to priority * 0.2.
+        try:
+            states_by_id: dict[str, Any] = {
+                s.neuron_id: s for s in await self._storage.get_all_neuron_states()
+            }
+            use_prefetched_states = True
+        except Exception:
+            _logger.debug(
+                "LIFECYCLE: get_all_neuron_states failed; falling back to a batch fetch",
+                exc_info=True,
+            )
+            states_by_id = {}
+            use_prefetched_states = False
 
-            access_count: int = int(neuron.metadata.get("access_frequency", 0))
+        # Fallback cache. `neurons` is already the whole pass — the paging above
+        # accumulates into one list — so this is a single get_neuron_states_batch
+        # over every id, not one per page, and it is populated on the first miss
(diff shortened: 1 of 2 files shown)


acidkill (approved):
Reviewed the full diff. LIFECYCLE now reads access_frequency/last_activated from NeuronState with a one-shot batch fallback; the fallback path is asserted to be a single query. Green CI.

PR #213, opened 2026-09-05
fix(smem-hygiene): remember the gate's refusals, not only its saves

RobertSigmundsson:
## Summary

- The auto-capture hooks now record the write gate's refusals in the idempotency ledger, not only its acceptances, so a rejected fragment is judged once per session instead of on every Stop.
- `PreCompact` honours a refusal `Stop` already recorded, and records its own.
- Adds ten tests, including three that count gate invocations rather than trusting a save count.

## Why

The ledger from #115 remembered what the gate accepted. A refusal left no trace at all, so the next Stop in the same session re-extracted the same fragment, re-encoded it, and put it to the gate again — which, being deterministic, refused it again. `PreCompact` then repeated the exercise from the top. Nothing crashes and nothing corrupts; the gate simply does the same homework on the same content forever, which is a shame in a mechanism that exists to stop precisely that.

Measured on one live brain over twenty-four hours on 2026-08-08: 499 auto decisions carried just 134 distinct contents, one fragment having been judged thirty-six times. That inflates the gate's denominator about 3.7-fold and makes its observed accept rate look several times worse than it is — 0.39% counted over rows against roughly 1.4% over distinct content. The wasted encoding is the smaller half of the cost; the misleading acceptance statistic is the half that gets acted upon.

## Changes

- `hooks/capture_state.py`: vocabulary for recording a refusal, keyed so that it is scoped to the threshold that produced it. A refusal recorded under one threshold does not suppress a re-judgement after the threshold changes, whilst an acceptance survives such a change.
- `hooks/stop.py`: the main capture loop records the refusal alongside the outcomes it already recorded.
- `hooks/pre_compact.py`: honours a refusal recorded by `Stop` rather than re-deciding it, and records its own refusals the same way.
- `tests/unit/test_capture_state.py` (+6) and `tests/unit/test_hook_capture_idempotency.py` (+4).

## One rough edge, disclosed rather than hidden

When every detected fragment was duplicate-filtered, the run reports the existing `duplicate content already captured` status and does not fall through to a session summary. Reusing that message is slightly imprecise — the two situations are not identical — but replacing an accurate duplicate status with a brand-new summary save would be the worse outcome, and a separate counter looked like more machinery than the distinction earns. Happy to split it if you would rather the two were distinguishable.

The module docstring in `test_hook_capture_idempotency.py` also loses a stale absolute path from a developer's machine. The file is being edited here anyway, so it seemed a good moment.

## Test plan

- [x] `pytest tests/unit/test_hook_capture_idempotency.py tests/unit/test_capture_state.py` — 43 passed.
- [x] Differential control, and the reason it is reported in two layers: an earlier shape of `test_pre_compact_honours_refusal_recorded_by_stop` **passed with the fix entirely removed**, because `PreCompact` never reached the gate on that input and the assertion could not distinguish that from success. The tests now count gate invocations and assert the gate *was* called before asserting it was not called again.
  - Reverting only `hooks/pre_compact.py`: 2 failed, 14 passed — `test_pre_compact_honours_refusal_recorded_by_stop` and `test_pre_compact_records_its_own_refusal`.
  - Reverting `hooks/pre_compact.py` and `hooks/stop.py`: 3 failed, 40 passed — the two above plus `test_stop_hook_does_not_rejudge_rejected_content`.
  - Each test is therefore pinned to the file whose hunk it covers.
- [x] Line coverage of the change itself, rather than of the modules: every executable line this PR adds to `capture_state.py`, `pre_compact.py` and `stop.py` is executed by the tests above — 12 statements, 0 uncovered. (The modules' own totals are low because they are large hook entry points with a great deal of unrelated code.)
- [x] `pytest tests/ -m "not stress" -n 4` against a live SurrealDB v3.2.0 — 7293 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus exactly the ten tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] Coverage under the CI gate: 72.44%, against 72.36% on `main`.

## Related issues

None filed. #115 introduced the ledger this extends; the refusal half was simply never part of it.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/hooks/capture_state.py
+++ b/src/surreal_memory/hooks/capture_state.py
@@ -83,6 +83,26 @@ def content_key(content: str) -> str:
     return hashlib.md5(norm.encode("utf-8")).hexdigest()
 
 
+def rejected_key(ck: str, min_score: int) -> str:
+    """Key marking a candidate the write gate REJECTED at ``min_score``.
+
+    Rejected candidates were never marked as seen, so every Stop/PreCompact
+    re-submitted them to the gate for as long as the transcript tail carried
+    them. Measured 2026-08-08: 499 auto decisions in 24h held only 134 distinct
+    contents (one fragment judged 36 times), inflating the gate's denominator
+    ~3.7x and making the observed accept rate look several times worse than it
+    was (0.39% on rows vs ~1.4% on distinct content).
+
+    Namespaced by the threshold **on purpose**: marking a rejection as a plain
+    seen-key would silence that content for the rest of the session even if the
+    threshold were lowered afterwards -- trading duplicate noise for silent
+    loss, which is the worse failure. Because the key embeds the score it was
+    judged against, changing ``auto_capture_min_score`` stops matching these
+    keys and every previously-rejected candidate gets a fresh hearing.
+    """
+    return f"rej{min_score}:{ck}"
+
+
 def _load_all() -> dict[str, Any]:
     path = _state_path()
     if not path.exists():
--- a/src/surreal_memory/hooks/pre_compact.py
+++ b/src/surreal_memory/hooks/pre_compact.py
@@ -129,7 +129,13 @@ async def flush_text(text: str, project_name: str | None = None) -> dict[str, An
     from surreal_memory.core.memory_types import MemoryType, Priority, TypedMemory
     from surreal_memory.engine.dedup.factory import build_dedup_pipeline
     from surreal_memory.engine.encoder import MemoryEncoder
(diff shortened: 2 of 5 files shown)


acidkill (approved):
Reviewed the full diff. Rejection keys namespaced by threshold are the right shape — remembered refusals without silencing content if auto_capture_min_score is later lowered; shared seen-set across Stop and PreCompact kept consistent.

PR #214, opened 2026-09-05
fix(mcp): smem_edit recomputes expires_at when the type changes, without reviving tombstones or ephemerals

RobertSigmundsson:
## Summary

- `smem_edit` now recomputes `expires_at` from the new type's default when the memory type changes, instead of leaving the old type's TTL in place.
- The recompute is skipped for soft-deleted and ephemeral memories, both of which carry an expiry that means something and must survive a type change.
- Adds four tests: one per direction of the recompute, one per guard.

## Why

`_edit`'s type-change branch swapped `memory_type` and left the old TTL untouched. `DEFAULT_EXPIRY_DAYS` is consulted only at creation time, through `remember_handler`'s `expires_in_days`, and `TypedMemory` is a frozen dataclass with no `__post_init__`, so nothing recomputed it afterwards either.

The result was wrong in both directions. A `DECISION` (90 days by default) edited to `FACT` (no default expiry) still expired roughly ninety days out, so a fact quietly vanished on the decision's clock. Going the other way, a `FACT` edited to `TODO` or `ERROR` (30 days each) never acquired a finite expiry and persisted indefinitely.

The fix reads `DEFAULT_EXPIRY_DAYS` for the new type after the swap and either clears `expires_at` or calls `TypedMemory.extend_expiry` — a helper that already existed on the dataclass and, until now, had no caller anywhere in `mcp/`.

## Two memories the recompute must not touch

An unconditional recompute is a data-integrity bug of its own, so it is guarded twice:

- **A soft-deleted memory.** `_forget(hard=False)` tombstones by setting `expires_at=utcnow()`. Recomputing from the new type's default would hand a deliberately forgotten memory an open-ended life — a `FACT` edit would make it immortal. The guard skips any record whose `expires_at` is at or before now.
- **An ephemeral memory.** `remember_handler` gives these a one-day TTL and flags the anchor neuron `ephemeral=True`. Clearing that expiry would turn a memory documented as "auto-expires after 24h, never synced" into a permanent one. The guard reads the anchor neuron and skips the recompute when the flag is set.

In both cases the type swap still applies; only the TTL recompute is skipped. `_forget` is untouched and remains the only other writer of `expires_at`.

## Changes

- `mcp/lifecycle_handler.py`: import `DEFAULT_EXPIRY_DAYS`; after the type swap, compute the tombstone and ephemeral conditions and, when neither holds, clear or extend the expiry.
- `tests/unit/test_edit_forget.py`: four tests, capturing the `TypedMemory` handed to `storage.update_typed_memory` — the interface between `_edit` and persistence, rather than any downstream effect. `test_edit_type_change_recomputes_ttl` (DECISION to FACT clears it), `test_edit_type_change_picks_up_finite_expiry` (FACT to TODO gains 30 days), `test_edit_type_change_does_not_resurrect_soft_deleted`, and `test_edit_type_change_preserves_ephemeral_ttl`.

## Not touched, on purpose

No new argument is added to `smem_edit`. Exposing `expires_at` explicitly would be the more flexible answer, but it is a schema change and a decision for you rather than for a bug fix. If you would rather have that, this PR is easy to rebase into it.

## Test plan

- [x] `pytest tests/unit/test_edit_forget.py` — 19 passed, against 15 on `main`; the delta is exactly the four tests added here.
- [x] Differential control, run twice, because the two halves of this change need separate proof:
  - Reverting **the whole hunk** to `main` and keeping the tests: 2 failed, 17 passed — the two *direction* tests fail, and the two guard tests pass, because with no recompute at all there is nothing for a guard to prevent.
  - Reverting **only the guards**, keeping the recompute: 2 failed, 17 passed — and this time it is exactly the two *guard* tests. The tombstone test sees its expiry resurrected; the ephemeral test sees its expiry cleared.
  - Neither pair can pass by accident, which is the property the first version of this change lacked.
- [x] `pytest tests/ -m "not stress" -n 4` against a live SurrealDB v3.2.0 — 7287 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the four tests here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] Coverage under the CI gate: 72.39%, against 72.36% on `main`.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/mcp/lifecycle_handler.py
+++ b/src/surreal_memory/mcp/lifecycle_handler.py
@@ -6,6 +6,7 @@ import logging
 from typing import TYPE_CHECKING, Any
 
 from surreal_memory.core.memory_types import (
+    DEFAULT_EXPIRY_DAYS,
     MemoryTier,
     MemoryType,
     Priority,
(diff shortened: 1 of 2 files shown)


acidkill (approved):
Reviewed the full diff. Type change recomputes expires_at from DEFAULT_EXPIRY_DAYS[new_type]; the tombstone/ephemeral carve-outs prevent resurrecting soft-deleted memories or making ephemerals immortal. Both directions tested.

Issue #110, opened 2026-07-30
Test suite overwrites the user's real claude_desktop_config.json


RobertSigmundsson:
## Summary

Running the unit test suite silently rewrites the user's real Claude Desktop MCP configuration at
`~/.config/Claude/claude_desktop_config.json`. Two tests call `run_full_setup()` without isolating
`Path.home()`, so production code that is *designed* to write the user's config runs against the
real home directory. The values written are taken from the pytest process environment, which other
tests have already polluted — so the user ends up with a config pointing at a dead endpoint.

This is not theoretical. It broke MCP for every Claude Code session on my machine and took two
debugging sessions to trace, because the failure surfaces as `-32000 Tool 'X' failed unexpectedly`
in a *different* application, hours later.

## Impact

- **Silent destruction of user configuration.** No prompt, no dry-run, no warning.
- The overwrite is invisible until Claude Desktop next re-initialises its MCP servers (on restart
  or account switch), which can be hours after the test run — making the two events look unrelated.
- A `.json.bak` is created, which is the only reason the original was recoverable.

### 1. Missing `Path.home()` isolation (the write)

```
tests/unit/test_full_setup.py::TestRunFullSetup::test_full_flow
tests/unit/test_full_setup.py::TestRunFullSetupSkipEmbeddings::test_skip_embeddings_does_not_prompt
```

Both call `run_full_setup()` with MCP setup enabled. Chain:

```
run_full_setup()
  → src/surreal_memory/cli/full_setup.py:351   setup_mcp_claude_desktop()
  → src/surreal_memory/cli/setup.py:273        setup_mcp_claude_desktop()
  → src/surreal_memory/cli/setup.py:251        _claude_desktop_config_path()  →  Path.home()/".config"/"Claude"/...
```

Every other test in the `setup_mcp_*` family correctly does
`patch("surreal_memory.cli.setup.Path.home", return_value=tmp_path)`. These two do not.
`test_skip_mcp` is unaffected because it passes `skip_mcp=True`.

### 2. Environment pollution across tests (the values)

```
tests/unit/test_bulk_batch_paths.py:77-81
tests/unit/test_batch_writers.py:34-38
```

Both use `os.environ.setdefault(...)` rather than `monkeypatch.setenv(...)`, so they mutate the
environment of the whole pytest process for the remainder of the session. `setup_mcp_claude_desktop()`
writes `build_mcp_env()` → `SurrealSettings.from_env()`, i.e. whatever is in the environment at
[...]

PR #215, opened 2026-09-05
fix(cli): the update check honours SURREAL_MEMORY_NO_UPDATE_CHECK, and the suite sets it

RobertSigmundsson:
## Summary

- `run_update_check_background()` gains an opt-out: it short-circuits when `SURREAL_MEMORY_NO_UPDATE_CHECK` is truthy.
- `tests/conftest.py` sets it for the session, so a full test run makes no calls to PyPI.
- Truthiness follows `unified_config._env_truthy`, the convention this repo already uses everywhere else.
- Adds nineteen tests over the accepted spellings, the rejected ones, and the unset case.

## Why

The update check had no opt-out at all. Most CLI invocations start a daemon thread that queries PyPI — the callback skips a short list of commands and `--json`, but not the several hundred invocations a full test run makes. On a machine with egress that is real outbound traffic from a unit test suite; on a machine without one it is the connection timeout instead.

It also feeds a flake, though not in the way it first appears. The banner goes to stderr, deliberately, with a comment in the code explaining that it must not break a piped `smem recall … | jq`. But `CliRunner` merges stderr into `result.output`, and the CLI tests parse that as JSON — so the banner lands in a different test's parse on each run.

Measured with `strace -f -e trace=connect`, counting connections to anything other than loopback, with a clean `HOME` so the 24-hour cache cannot mask the result:

| tree | `SURREAL_MEMORY_NO_UPDATE_CHECK` | non-loopback connections |
|---|---|---|
| `main` | unset | 10 |
| `main` | `1` | 10 — the variable does not exist yet |
| this branch | unset | 10 — the check still works |
| this branch | `1` | 0 |
| this branch | `on`, `TRUE` | 0 |
| this branch | `off`, `0` | 10 |

The third row is the one that matters as a control: the opt-out suppresses the check, it does not break it.

## Why this truthiness and not the obvious one

The obvious implementation — "any non-empty value disables it" — would give `SURREAL_MEMORY_NO_UPDATE_CHECK=off` the effect of switching the check *off*, which is the opposite of what anyone typing that would expect, and it would be a fourth convention in a repo that already has one.

So the gate calls `unified_config._env_truthy`, the same helper behind `SURREAL_MEMORY_EMBEDDING_ENABLED`, `SURREAL_MEMORY_SYNC_ENABLED`, `SURREAL_MEMORY_SYNC_AUTO` and the three `SURREAL_MEMORY_REASONING_*` switches: `1/true/yes/on`, case-insensitive, and everything else — `off`, `no`, `0`, `false` included — leaves the check running. The last two rows of the table above are that decision, measured.

## Changes

- `cli/update_check.py`: the short-circuit and a docstring recording the convention and why the test suite sets the variable.
- `tests/conftest.py`: `_isolated_home_dir` sets `SURREAL_MEMORY_NO_UPDATE_CHECK=1` alongside the `$HOME` redirect it already installs. Issue #110 and PR #121 closed the write side of that fixture; this closes the network side. A test that needs the real code path can `monkeypatch.delenv` it, and `tests/unit/test_update_check_env_gate.py` does exactly that.
- `tests/unit/test_update_check_env_gate.py` (new).

## Test plan

- [x] `pytest tests/unit/test_update_check_env_gate.py` — 19 passed.
- [x] The `strace` table above: measured on both trees with a fresh `HOME` each time. The first attempt at this probe reported zero connections everywhere, because the check caches its answer for 24 hours in `~/.surrealmemory/.update_check` and the real `HOME` had a warm cache — worth mentioning in case anyone reproduces it and wonders why nothing happens.
- [x] `pytest tests/ -m "not stress" -n 4` against a live SurrealDB v3.2.0 — 7302 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus exactly the nineteen tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] Coverage under the CI gate: 72.30%, against 72.36% on `main`.

## On the flake

The JSON-parsing flake in the CLI tests comes from the same banner reaching `result.output` through `CliRunner`, and setting the variable for the suite removes the banner. I am reporting that as a consequence rather than as this PR's proof: a green run is not by itself evidence that a flake is gone.

## Verified by

@RobertSigmundsson


acidkill (approved):
Reviewed the full diff. Env gate follows the repo's canonical _env_truthy convention, conftest sets it session-wide, and falsy values (off/no/false) deliberately do not disable — an env var that looks like a switch never silently disables the check.

PR #217, opened 2026-09-05
fix(server): PUT /memory/neurons refreshes content_hash and the embedding on a content change

RobertSigmundsson:
## Summary

- `PUT /memory/neurons/{id}` now refreshes `content_hash` and the embedding when the content changes, instead of re-saving the old ones against the new text.
- It carries the internal `_embedding` key forward when the same request also sends `metadata`, which is the case where refreshing the hash alone would leave the vector stranded.
- A metadata-only PUT, or one whose content is unchanged, stays a single write with no provider call.
- Adds five tests covering both halves and the 404.

## Why

The route built `replace(neuron, **updates)` with the new content and handed it to `update_neuron`, which derives `content_hash` from the object and the vector from `metadata["_embedding"]`. Both therefore continued to describe the text that had just been replaced: the memory stayed findable by content that no longer existed, and near-duplicate detection compared it against a fingerprint of the old text.

This is the class #166 fixed for `smem_edit` and #193 fixed for the engine paths. #193 explicitly left this route out as a separate surface with its own test story; this is that follow-up.

## The half that is easy to miss

`content_refreshed` re-embeds only when `metadata["_embedding"]` is present — that is how it tells "this neuron has a vector to refresh" apart from "this one never had one". A PUT that sends `metadata` alongside a content change replaces the whole dict, so the key is gone before the helper ever sees it. The hash gets refreshed, the vector silently does not, and nothing is logged.

That failure is invisible from the outside, which is why it is worth stating how it was measured rather than asserted. Against a live server and a live SurrealDB v3.2.0, with a neuron seeded with a 3072-dimension vector and matching `_embedding`:

| PUT | this branch | the same route without the carry-forward hunk |
|---|---|---|
| content only | refresh path entered | refresh path entered |
| **content + metadata** | **refresh path entered** | **not entered** |
| `content_hash`, either request | changes | **changes** |
| server log, content + metadata | `WARNING … Content of neuron … changed but the embeddings could not be recomputed - the stored vectors still describe the old content; run 'smem reindex --all' to repair them.` | silent |

Over the same pair of requests that is two entries against one. The content-only PUT behaves identically on both — it always reached the helper — so the whole difference sits in the row in bold. The hash changes either way, which is precisely how an earlier version of this change looked correct whilst half of it did nothing: measuring the hash proves nothing at all about the vector.

## Changes

- `server/routes/memory.py`: pop `content` out of the update set, apply the rest via `replace`, and call `content_refreshed` only when the content actually differs. Before that call, if the stored neuron had an `_embedding` and the incoming `metadata` does not carry one, the key is carried forward so the helper can act on it. A caller who supplies their own `_embedding` keeps it.
- `tests/unit/test_server_update_neuron_content_refresh.py` (new): content-only, content with metadata, metadata-only, unchanged content, and the 404.

## Test plan

- [x] `pytest tests/unit/test_server_update_neuron_content_refresh.py` — 5 passed.
- [x] Differential control in two layers, because the change has two halves:
  - Reverting the whole route to `main`: all 5 fail.
  - Reverting only the `_embedding` carry-forward and keeping the rest: exactly 1 fails, `test_content_and_metadata_put_preserves_embedding_for_refresh`. The other four still pass, which is what makes that test worth having.
- [x] Live, against `smem serve` and a SurrealDB v3.2.0 instance: the table above, plus a metadata-only PUT and an unchanged-content PUT that both correctly skip the refresh, `404` for an unknown id, `400` for a `type` outside the enum, and `405` for `PATCH`.
- [x] `pytest tests/ -m "not stress" -n 4` — 7288 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the five tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.

One limit, stated rather than glossed. The live box had no embedding provider installed, so the *new vector value* was not measured end to end: the refresh is attempted and reports that it could not complete. What the live run establishes is that the attempt happens at all, which is the entire difference between the two versions. The unit tests do not close that gap either — they patch `content_refreshed` itself, so they prove the route hands the helper a neuron that still carries `_embedding`, not that the helper then recomputes the vector. The helper's own behaviour is covered where the helper lives. If you would rather see the recomputation end to end before merging, say so and I will run it against a box with a provider.

## Verified by

@RobertSigmundsson

### DIFF
--- a/src/surreal_memory/server/routes/memory.py
+++ b/src/surreal_memory/server/routes/memory.py
@@ -31,6 +31,7 @@ from surreal_memory.server.models import (
     SynapseUpdateRequest,
 )
 from surreal_memory.storage.base import NeuralStorage
+from surreal_memory.utils.content_refresh import content_refreshed
 
 router = APIRouter(
     prefix="/memory",
@@ -526,7 +527,29 @@ async def update_neuron(
     if request.metadata is not None:
         updates["metadata"] = request.metadata
 
-    updated = replace(neuron, **updates)
+    # Apply type/metadata first via replace(), then refresh content-derived
+    # fields (content_hash, embedding) only when content actually changed.
+    # Same helper as smem_edit (#166) and the engine paths (#193); a
+    # metadata-only PUT or a PUT with unchanged content stays a one-write op.
+    new_content = updates.pop("content", None)
+    updated = replace(neuron, **updates) if updates else neuron
+    if new_content is not None and new_content != updated.content:
+        # `_embedding` is an internal key (`utils/content_refresh.py` reads it
+        # to decide whether to re-embed). A PUT that ships `metadata` alongside
+        # a content change replaces the whole dict, so the pre-edit `_embedding`
+        # would be gone before the helper ever saw it — and the old vector
+        # would stay in the row, describing text that no longer exists. Carry
+        # the internal key forward (unless the caller explicitly provided
+        # one) so content_refreshed can refresh it.
+        if (
+            neuron.metadata.get("_embedding") is not None
+            and updated.metadata.get("_embedding") is None
+        ):
+            updated = replace(
+                updated,
+                metadata={**updated.metadata, "_embedding": neuron.metadata["_embedding"]},
+            )
+        updated = await content_refreshed(storage, updated, new_content)
     await storage.update_neuron(updated)
 
     return NeuronResponse(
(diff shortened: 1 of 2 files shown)


acidkill (approved):
Reviewed the full diff. Same content_refreshed helper as smem_edit and the engine paths; the _embedding carry-forward on metadata-replacing PUTs is the detail that makes it correct. Guarded to fire only on real content change.

PR #218, opened 2026-09-05
fix(storage): change_log.payload needs FLEXIBLE, or every change row is dropped

RobertSigmundsson:
## Summary

- `change_log.payload` is declared `FLEXIBLE`, so a nested payload is stored instead of rejected. Without it, no change-log row was written at all on the SurrealDB backend.
- The declaration uses `OVERWRITE`, which is what carries the fix to databases that already exist.
- Both change-log writers now report a failed insert instead of swallowing it.
- Adds a live test: one `add_neuron` must leave one replayable row.

## Why

`change_log` is `SCHEMAFULL` and `payload` was `option<object>` with no `FLEXIBLE`. SurrealDB rejects a nested object against such a field outright:

```
CREATE change_log SET brain_id = 'b1', payload = { content: 'hello', tags: ['x'] };
-- Found field 'payload.content', but no such field exists for table 'change_log'
```

`payload = NONE` is accepted, which is why the table looked healthy. Every neuron, synapse and fiber write builds a nested payload, so on this backend the insert failed every time and the row never landed — sync had nothing to replay and nothing anywhere said so. `_record_change_internal` swallowed the error with `except Exception: pass`; `_record_changes_bulk` logged it at debug.

`decay_pass.config_snapshot`, the table directly above in the same file, already declares `FLEXIBLE` for exactly this reason.

## Why `OVERWRITE`, and why that is not a detail

A plain re-`DEFINE` of a field that already exists raises `The field 'payload' already exists`, and `ensure_schema` swallows "already exists" at debug by design — its own docstring records the same effect for `DEFINE INDEX`, noting that a bare `DEFINE` therefore does not change an existing object. So editing the line without `OVERWRITE` would have fixed new databases and left every existing one dropping change rows exactly as before, whilst the release notes said it was fixed.

Measured on SurrealDB 3.2.0:

| statement | field already exists? | result |
|---|---|---|
| `DEFINE FIELD payload … FLEXIBLE` | yes | `The field 'payload' already exists` — no change |
| `DEFINE FIELD OVERWRITE payload … FLEXIBLE` | yes | applied; nested payload then accepted |
| `DEFINE FIELD OVERWRITE payload … FLEXIBLE` | no | applied; nested payload accepted |

It stays in `SCHEMA_SQL` rather than behind a `SCHEMA_VERSION` bump, for the reason the file already gives for `training_files` and `watch_state`: `ensure_schema` is idempotent and runs on every `initialize()`. Bumping the version would re-key the `(9, 10)` entry in `MIGRATIONS` off `TARGET_VERSION` and strand databases still on v9. (`schema.py`'s own note still describes this in terms of an `(8, 9)` entry, which is the situation the registry comment at `migrations.py` has since pinned away — worth not repeating.)

## The writers

Both now report the failure, but not identically. `_record_changes_bulk` warns outright — it runs once per batch. `_record_change_internal` warns on the first failure in the process and drops to debug afterwards, because it sits on the hot write path; `_skip_change_log` exists precisely because of how often it runs, so an unconditional per-entity warning would trade a silent failure for a log flood. The fail-soft contract is unchanged in both: a change-log problem must never abort the entity write that already succeeded.

## Test plan

- [x] `pytest tests/unit/test_surrealdb_change_log_payload_live.py` — 1 passed against a live SurrealDB v3.2.0.
- [x] Differential control, in both halves. With the source reverted **and** the field reset to `option<object>` without `FLEXIBLE`, the test fails on `add_neuron left no change_log row at all`. Restoring the source and re-running `initialize()` re-applies `FLEXIBLE` to that same already-degraded database and the test passes again — which is the migration path being exercised rather than described.
- [x] The table above, measured directly against the running database.
- [x] `pytest tests/ -m "not stress" -n 4` — 7284 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the one test added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.
- [x] Coverage under the CI gate: 72.35%, against 72.36% on `main`.

## Related issues

None filed; this was found whilst writing up a different report about `get_changes_since`, whose reproduction recipe could not work precisely because nothing was reaching `change_log` in the first place.

## A note for whoever merges second

`tests/unit/_surrealdb_live.py` conflicts textually with the record-id fix: both append a brain name to `LIVE_TEST_BRAIN_NAMES` immediately after the same entry. The resolution is to keep both lines; I will rebase whichever of us lands second.

## Verified by

@RobertSigmundsson


acidkill (approved):
Reviewed the full diff. DEFINE FIELD OVERWRITE ... FLEXIBLE is the correct spelling — it fixes fresh databases and migrates existing ones without breaking ensure_schema idempotence; the once-per-process warning replaces a silent swallow on the hot path.

acidkill:
Two follow-up commits pushed to resolve the batch's cumulative drift, so this PR can land green:

1. Merge of `main` — `tests/unit/_surrealdb_live.py` add/add conflict with #208 resolved as the union of both brain-name entries.
2. `scripts/sync_refs.py --fix` — the 20-fix batch pushed the collected test count past 7400; the five scattered '7300+ tests' references (README, README.es-ES, ROADMAP, FAQ, promo) are regenerated. Mechanical only.

PR #219, opened 2026-09-05
fix(storage): the SurrealDB change readers return the record type sync consumes

RobertSigmundsson:
## Summary

- `SurrealDBStorage.get_changes_since` and `get_unsynced_changes` return `ChangeEntry`, the record type `SyncEngine` actually reads and the in-memory backend already returns, instead of `SyncChange`.
- Adds a live test that performs the two attribute accesses `SyncEngine` performs, rather than asserting on types alone.

## Why

`SyncEngine` reads `change.id` and calls `change.changed_at.isoformat()` on whatever these two readers hand back. `SyncChange` has no `id`, and its `changed_at` is a `str`. So on the SurrealDB backend both sync paths raised before sending anything. Reproduced against a live database:

```
'SyncChange' object has no attribute 'id'
'str' object has no attribute 'isoformat'
```

The first fires in `create_sync_request`, the second in `process_sync_request` — the device side and the hub side respectively, so neither direction of a sync survives.

The abstract methods in `storage/base.py` are annotated `-> list[Any]`, which is how two backends came to return different record types from the same call with nothing complaining. `test_change_log.py` already pins `ChangeEntry` for the in-memory backend; there was no equivalent on this side.

## Which side to change

The SurrealDB side, for two reasons. `ChangeEntry` carries `brain_id` and `synced`, and `SyncChange` has nowhere to put them — converting the other way would lose information the caller may want. And `SyncEngine` builds `SyncChange` objects from these results itself, a few lines later; it is the wire format, not the storage format.

## Changes

- `storage/surrealdb/store.py`: a shared `_row_to_change_entry` builds the record; both readers use it. `sequence` comes back from the SDK as an `int` and `changed_at` as a `datetime` — both measured — but the helper coerces anyway, since a row written by an older build may carry either, and falls back to the current time rather than raising on a value it cannot parse. A sync read should not be the thing that dies on one malformed row.
- `tests/unit/test_surrealdb_sync_change_entry_live.py` (new): asserts the returned objects are `ChangeEntry`, and then performs `change.id` and `change.changed_at.isoformat()` for real.

## This is stacked on the `change_log` FLEXIBLE fix, and has to be

On a database whose `change_log.payload` is still declared without `FLEXIBLE`, this bug is **unreachable**: no change-log row is ever written, so the readers hand back an empty list and nothing raises. That is the reason the obvious reproduction for this one does not work on a stock database — and it is also why this branch carries that fix as its first commit rather than pretending to be independent.

Measured, because I got this wrong first time round: on a **fresh** namespace the live tests here fail without that schema change, and this repository's `Integration (SurrealDB)` job starts a fresh v3.2.0 instance for every run. So a version of this PR that did not carry the schema fix would arrive with a red CI — which is the first item on the AI-slop list in `CONTRIBUTING.md`, and rightly so.

The diff therefore shows two commits. Once the schema fix merges, this branch rebases to one and the diff shrinks to the reader change alone; the two also conflict textually in `storage/surrealdb/store.py` and `tests/unit/_surrealdb_live.py`, where both add a symbol immediately after the same anchor, and that conflict is already resolved here. Happy to reopen this as a single-commit PR afterwards if you would rather review it that way.

## Test plan

- [x] `pytest tests/unit/test_surrealdb_sync_change_entry_live.py` — 2 passed against a live SurrealDB v3.2.0, on a **freshly created** namespace rather than a reused one.
- [x] With `store.py` reverted to `main` and the tests kept, both fail — the first on `AssertionError: got ['SyncChange']`.
- [x] `pytest tests/unit/test_change_log.py` — 23 passed, unchanged; the in-memory contract this aligns with is untouched.
- [x] `pytest tests/ --timeout=120 -m "not stress"` — the `Integration (SurrealDB)` job's own command, serial, on a freshly created namespace — 7288 passed, 48 skipped, 1 xfailed, which is `main`'s 7285 plus three tests: the two added here and the one that comes with the schema commit this is stacked on. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail under `-n` on this branch and on `main` alike — they collide with one another on a shared database — and pass when the suite runs serially, which is how the `Integration` job runs it.
- [x] Coverage stays above the CI gate; the branch adds tests and no production code beyond the two commits described.

## Verified by

@RobertSigmundsson


acidkill:
Pushed a merge of `main` onto the branch: `store.py` keeps the `_row_to_change_entry` helper over the (empty) main side, and `_surrealdb_live.py` resolves as the union of the brain-name entries with #208/#218. ruff and both docs gates verified locally.

acidkill (approved):
Reviewed the full diff. ChangeEntry is what SyncEngine actually consumes (id + datetime changed_at) — the SyncChange returns made both sync paths AttributeError on the only production backend; coercions keep older-build rows readable. Merge-of-main pushed to clear the stacked #218/#208 textual overlaps (union in the brain-name set).

PR #220, opened 2026-09-05
fix(storage): SURREALDB_AUTH_LEVEL, so a database-scoped user can sign in at all

RobertSigmundsson:
## Summary

- `SURREALDB_AUTH_LEVEL` selects the scope sign-in uses: `root` (the default, and the shape sent today), `namespace`, or `database`.
- A user defined `ON DATABASE` can now authenticate at all; before, no configuration allowed it.
- `AUTH_HINT` mentions the level, since it previously blamed the password in exactly the case where the password is right.
- Adds ten unit tests and two live ones that drive the real `signin`.

## Why

All four sign-in call sites sent `{"username": …, "password": …}` and nothing else. SurrealDB accepts that shape only for a root user; a user defined `ON DATABASE` is rejected outright. So an installation following the ordinary advice — create a scoped user rather than handing the application root — could not connect, and the error named authentication without naming the scope.

Measured against SurrealDB 3.2.0, with a user created by `DEFINE USER … ON DATABASE`:

| identity | `{user,pass}` | `{user,pass,ns}` | `{user,pass,ns,db}` |
|---|---|---|---|
| root | **accepted** | rejected | rejected |
| `ON NAMESPACE` | rejected | **accepted** | rejected |
| `ON DATABASE` | rejected | rejected | **accepted** |

Each identity accepts exactly one shape and rejects the other two — which is why there are three levels rather than a boolean, and why none of them can be the unconditional default. That is why this cannot be fixed by adding the namespace and database unconditionally: `DEFAULT_USER` is `root`, so doing that would reject every stock installation — trading an uncommon broken configuration for the common working one.

## Changes

- `storage/surrealdb/connection.py`: `AUTH_LEVELS`, `DEFAULT_AUTH_LEVEL = "root"`, an `auth_level` field on `SurrealSettings` read from `SURREALDB_AUTH_LEVEL`, and `signin_payload`, which builds the payload for a level in one place.
- `storage/surrealdb/store.py` and `cli/doctor.py`: all four sites use it.
- `AUTH_HINT` gains a sentence naming the level as a cause.
- `docs/getting-started/installation.md`: the new variable in the environment table.

An unrecognised level falls back to `root` rather than raising. A typo in a connection setting should not be something you have to fix before you can connect at all, and silently changing the scope would be worse than ignoring the typo.

`SurrealDBStorage` carries the level as a class-level default as well as an instance attribute. Several tests build the store with `__new__` and seed only the attributes they drive; an instance-only assignment left the reconnect path raising `AttributeError` instead of signing in the way a stock installation does. The full suite caught it, which seemed worth mentioning since the class attribute otherwise looks redundant.

## Test plan

- [x] `pytest tests/unit/test_signin_auth_level.py` — 10 passed: each payload shape, the default, case- and whitespace-insensitivity, and the fallback.
- [x] `pytest tests/unit/test_signin_auth_level_live.py` — 2 passed against a live SurrealDB v3.2.0. They create a user `ON DATABASE` and drive the real `signin`, asserting each identity authenticates at its own level and **not** at the other's. The negative half is the part that matters: it is what makes an unconditional change visibly wrong rather than merely untested.
- [x] With the three source files reverted to `main` and the tests kept, both files fail at collection — `signin_payload` does not exist there, which is the shape of the change.
- [x] `pytest tests/ -m "not stress" -n 4` — 7295 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the twelve tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.

## One thing left alone

`doctor.py` signs in without calling `use()` afterwards on one of its two probes. That is unrelated to the level and I have not changed it — it only asks for the server version, which needs no namespace. Mentioning it because it looks like an omission next to this diff and is not one.

## Verified by

@RobertSigmundsson


acidkill:
Pushed a merge of `main` onto the branch (no textual conflicts) plus the regenerated `docs/reference/config.md` footer — with #215's env var also landed, the generator count is 53, not 52.

acidkill (approved):
Reviewed the full diff. Three signin payload shapes (root/namespace/database) match SurrealDB's auth model, the default stays root so stock installations are untouched, and the doctor/migration probes use the same helper. Docs regenerated to 53 env vars after the #215 union.

Issue #80, opened 2026-07-17
Auto-capture: no idempotency key (Stop/PreCompact re-capture loop) + PreCompact/task_context hooks still bypass dedup


RobertSigmundsson:
**Context:** surreal-memory in production for our agent swarm (Uruboros), via stdio MCP + Claude Code Stop/PreCompact auto-capture. Open cards; fix ready on our fork.

**Since #69** the Stop hook has embedding-based semantic dedup — good. **Two gaps remain:**

1. **No idempotency key at the source.** The Stop hook fires after every assistant turn and PreCompact on each compaction; both re-read an overlapping transcript tail and re-run `analyze_text_for_memories`, re-encoding identical fragments across turns. The embedding dedup in `stop.py` operates within a single candidate batch, so it does not stop the cross-turn re-capture loop (the same fragment persisted last turn is re-encoded this turn).
2. **PreCompact and task_context hooks have no dedup at all** — `pre_compact.py` and `task_context.py` still construct `MemoryEncoder` without any dedup pipeline.

**Impact:** the noisiest write path (auto-captured session summaries) accumulates near-duplicates — the memory-poisoning failure mode — on any host using the default Stop/PreCompact integration.

**Fix we're offering (PR-ready on our fork, will align after our 2.11.0 rebase):** an idempotency key at the source to break the re-capture loop; dedup wired into pre_compact + task_context; exact-content candidate lookup so exact dups can't escape the word-search window. On our deployment this dropped `content_hash=0` fresh neurons from ~64% to ~0.25%.

acidkill:
Re-verified against current `main`, not just the description — both gaps are still open:

- `_embedding_dedup` exists only in `hooks/stop.py` (line 221). `hooks/pre_compact.py` and `hooks/task_context.py` still construct `MemoryEncoder` with no dedup pipeline.
- No idempotency key anywhere in the write path (`git grep idempot` on `main` is empty).

**What changed today that you should rebase onto.** #93 and #94 both landed and both touch this area from the other end:

- #93 adds write-gate modes `off | shadow | enforce` plus a per-intent `auto_capture_mode`, so the gate can be enforced on `intent=auto` while interactive writes stay in shadow. It also touches `hooks/pre_compact.py` and `hooks/stop.py`, so expect conflicts there.
- #94 adds an empirical machine-noise denylist to `quality_scorer`.

[...]

PR #221, opened 2026-09-05
fix(smem-hygiene): default the Stop hook's session-summary fallback to off

RobertSigmundsson:
## Summary

- The Stop hook's session-summary fallback becomes opt-in: `auto.capture_session_summary`, defaulting to `false`.
- The fallback is a truncation, not a summary; on a stock installation the write gate is off, so what it produces is stored rather than filtered.
- Adds six tests over the default, the round trip, and the behaviour on both settings.

## Why

When pattern extraction finds nothing in a transcript, the Stop hook manufactures a candidate from the last few lines and offers that to the write gate. On a real chat transcript those lines are harness markers and half-sentences, not knowledge — so the gate rejects them, and the work of extracting, encoding and judging them is repeated on every turn that produces no patterns.

The repository already says as much in two places, which is what makes this a default worth changing rather than a preference:

- `quality_scorer.py`'s `_NOISE_PREFIXES` opens with the literal `"session activity:"`, above a comment describing it as an *"Empirical denylist from gate_decision false-accepts (2026-07-23): 180x "Session activity:", 30x <summary>, …"*. The single most common false accept the gate had to learn to reject is the thing this fallback manufactures.
- `capture_state.py`'s module docstring calls duplicate `"Session activity"` entries *"the dominant source of memory poisoning"*.

Measured on one production brain over twenty-four hours: 96 of 158 write-gate rejections, a little over three in five, originated in this path alone — entries like `Session activity: <task-notification>`. That number is from a single brain and I would not lean on it alone; it agrees with the direction of the two notes above, which are the project's own.

## Changes

- `unified_config.py`: the new `auto.capture_session_summary` field, defaulting to `false`, with the reason recorded beside it.
- `hooks/stop.py`: the fallback runs only when the flag is set.
- `tests/unit/test_stop_hook_session_summary.py` (new) and `tests/unit/test_hook_capture_idempotency.py`, which turns the fallback back on explicitly.

## What this does not change

The gate itself. `WriteGateConfig` defaults to `enabled: False` and `mode: "off"`, so on a stock installation the fallback's output is not rejected — it is *stored*, as a `CONTEXT` memory. That is the case this default actually addresses, and it is the common one rather than the exception, which is worth being explicit about: the argument here is not "the gate wastes effort" but "on the defaults, nothing is stopping this".

Setting `capture_session_summary = true` restores the previous behaviour exactly. The #80 idempotency regression tests do that deliberately, since the defect they pin lives on this path and needs it switched on to be observable at all.

## Not touched, on purpose

The extraction helper keeps its name. Renaming it to say "tail" rather than "summary" would describe it better, and it has a single call site, so the rename would be small — but a rename is a different change from a default flip and belongs in its own commit rather than riding along.

## Test plan

- [x] `pytest tests/unit/test_stop_hook_session_summary.py` — 6 passed: the default is off, an absent key loads as off, an explicit `true` survives the config round trip, `save` emits the key, and the hook stores nothing for pattern-free text with the flag off and does store with it on.
- [x] `pytest tests/unit/test_hook_capture_idempotency.py` — 12 passed. Those tests turn the fallback back on explicitly, because the #80 defect they exist to pin is only reachable through it.
- [x] With `hooks/stop.py` and `unified_config.py` reverted to `main` and the tests kept, 5 of the 6 new tests fail.
- [x] `pytest tests/ -m "not stress" -n 4` — 7289 passed, 48 skipped, 1 xfailed, which is `main`'s 7283 plus the six tests added here. Two tests in `tests/unit/test_dashboard_brains_scope.py` fail on this branch and on `main` alike: they want a live database and collide with one another under `-n`. Both pass when that file is run on its own.

## Verified by

@RobertSigmundsson

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -63,6 +63,7 @@ Auto-capture configuration for MCP server.
 | `capture_insights` | `bool` | `true` |  |
 | `capture_preferences` | `bool` | `true` |  |
 | `min_confidence` | `float` | `0.7` |  |
+| `capture_session_summary` | `bool` | `false` |  |
 
 ## `[eternal]`
 
--- a/src/surreal_memory/hooks/stop.py
+++ b/src/surreal_memory/hooks/stop.py
@@ -501,12 +501,16 @@ async def capture_text(text: str, project_name: str | None = None) -> dict[str, 
                 logger.debug("Failed to save stop-hook memory", exc_info=True)
                 continue
 
-        # Always save a session summary if no patterns were detected. Skipped
+        # Fall back to a session summary if no patterns were detected. Skipped
         # when every detected fragment was duplicate-filtered above -- that
         # case already has an accurate "duplicate" status; falling through
         # here would silently replace it with a brand-new summary save.
+        #
+        # Off by default (auto.capture_session_summary): the "summary" is the last
+        # ~10 transcript lines verbatim, which on a real chat is harness markers and
+        # half-sentences, not knowledge. See AutoConfig for the measured numbers.
         summary_had_candidate = False
-        if not saved and not all_candidates_were_duplicates:
+        if config.auto.capture_session_summary and not saved and not all_candidates_were_duplicates:
             summary = _extract_session_summary(text)
             if summary:
                 summary_had_candidate = True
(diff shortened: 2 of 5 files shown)


acidkill (approved):
Reviewed the full diff. The measured 61%-of-rejections number justifies the default flip; both directions are pinned, the idempotency suite opts back in explicitly, and the subprocess racer gets config.toml rather than an in-process toggle. Note for the changelog: default behaviour change, opt-out is auto.capture_session_summary=true.

Issue #224, opened 2026-09-05
`Integration (SurrealDB)` can hang at teardown of `TestListBrainsScopeLive` until the 120 s timeout


RobertSigmundsson:
## Description

In a full serial run against a live SurrealDB — which is exactly what the `Integration (SurrealDB)`
job does — `tests/unit/test_dashboard_brains_scope.py::TestListBrainsScopeLive::
test_request_leaves_the_shared_singleton_on_the_active_brain` sometimes does not finish its
**teardown**. The test itself passes; what hangs is pytest-asyncio closing the function-scoped
event loop: `runner.close()` → `_cancel_all_tasks` → `loop.run_until_complete(gather(...))`
→ `selector.poll(...)` forever, until pytest-timeout fires at 120 s and reports the test as an
`error`. The run is otherwise green, so the job goes red on a test that asserted nothing wrong.

It also reproduces with the file on its own, less often: 22 isolated runs of the class, 3 of them
ending in the same teardown timeout, 2 of those on an untouched `main`.

## To Reproduce

1. Start a fresh `surrealdb/surrealdb:v3.2.0` (`start memory`), export `SURREALDB_URL`,
   `SURREALDB_USER`, `SURREALDB_PASS`, `SURREALDB_NS`, `SURREALDB_DB` as `ci.yml` does.
2. Cheap: `pytest tests/unit/test_dashboard_brains_scope.py --timeout=120`, repeated ~10 times on a
   fresh namespace each — roughly one run in seven ends `10 passed, 1 error in ~122 s`.
3. Or the job's own command, `pytest tests/ --timeout=120 -m "not stress"` — serial, no `-n`. A run
   that stays clear finishes in 86-165 s; one that hits this ends in 209-252 s, the difference being
   the single 120 s timeout.

## Expected Behavior

Teardown of the test completes in milliseconds, every run.

## Actual Behavior

Intermittently the loop never drains and pytest-timeout ends the teardown:

```
_ ERROR at teardown of TestListBrainsScopeLive.test_request_leaves_the_shared_singleton_on_the_active_brain _
.venv/…/pytest_asyncio/plugin.py:1052: in _scoped_runner
    runner.__exit__(None, None, None)
…/asyncio/runners.py:206: in _cancel_all_tasks
    loop.run_until_complete(tasks.gather(*to_cancel, return_exceptions=True))
…/selectors.py:468: in select
    fd_event_list = self._selector.poll(timeout, max_ev)
[...]

PR #225, opened 2026-09-05
fix(dashboard): hold the background diagnostics refresh, and drain it in the live tests

RobertSigmundsson:
## Summary

- `_schedule_grade_refresh` keeps its task only in a local variable, although its docstring says a
  strong reference is held until the task finishes. The loop keeps only a weak reference, so the
  refresh can be collected mid-flight and the stale grade is never replaced.
- `_GRADE_REFRESH_TASKS` now holds it, exactly as `_BRAINS_REFRESH_TASKS` already holds the brain
  refresh a few lines below.
- That reference also gives the live dashboard tests something to wait for, which removes an
  intermittent 120 s teardown timeout in `Integration (SurrealDB)`.

## Why

The two background refreshes were written together and guard the same hazard, but only one of them
keeps its task:

```python
task = asyncio.get_running_loop().create_task(_refresh())
task.add_done_callback(_done)          # nothing else holds `task`
```

against, in `_refresh_in_background`:

```python
task = asyncio.get_running_loop().create_task(_run())
# Strong reference: the event loop discards tasks nobody holds.
_BRAINS_REFRESH_TASKS.add(task)
```

The consequence in production is a diagnostics refresh that may never finish, leaving the stale
report in place until the next request schedules another one.

The consequence in CI is a red job on unrelated branches. `/api/dashboard/brains` serves stale data
and refreshes behind the request, so a task is still querying the store when a test returns.
pytest-asyncio tears the function-scoped loop down and cancels it, and occasionally the loop waits
on a cancelled receive until pytest-timeout fires:

```
_ ERROR at teardown of TestListBrainsScopeLive.test_request_leaves_the_shared_singleton_on_the_active_brain _
E   Failed: Timeout (>120.0s) from pytest-timeout.
= 7285 passed, 48 skipped, 32 deselected, 1 xfailed, 1 error in 218.45s (0:03:38) =
```

## Changes

- `src/surreal_memory/server/routes/dashboard_api.py`: `_GRADE_REFRESH_TASKS` holds each scheduled
  diagnostics refresh and drops it in a done callback.
- `tests/unit/test_dashboard_cache.py`: a test schedules a refresh that blocks, asserts the task is
  referenced outside the loop while it runs, and asserts the set is empty once it finishes.
- `tests/unit/test_dashboard_brains_scope.py`: the two live classes cancel and await both task sets
  at teardown, on the loop that owns them.

## Test plan

- [x] `pytest tests/unit/test_dashboard_cache.py -k held_until` against `main`'s production code —
      fails. With the module-level set present but the task not added to it, the same test fails on
      `assert set()`, so it pins the behaviour rather than the symbol.
- [x] The same test on this branch — `15 passed` for the file.
- [x] The two live classes run 40 times in a row against a live SurrealDB v3.2.0, a freshly created
      namespace each time and nothing else running against that database — 40 clean. Before the
      change the same loop produced 3 teardown timeouts in 22 runs.
- [x] `pytest tests/ --timeout=120 -m "not stress"` serial on a fresh namespace, the
      `Integration (SurrealDB)` command — 7286 passed, 48 skipped, 1 xfailed, in 96 s. The control
      run of `main` immediately afterwards, same database, hit the timeout: 7285 passed and 1 error
      in 215 s.
- [x] `ruff check src/ tests/` clean; `ruff format --check src/ tests/` reports 739 files already
      formatted.
- [ ] `CHANGELOG.md` untouched — left to the release entry.

## Related issues

Reported in #224.

### DIFF
--- a/src/surreal_memory/server/routes/dashboard_api.py
+++ b/src/surreal_memory/server/routes/dashboard_api.py
@@ -42,6 +42,7 @@ _GRADE_REFRESH_KEYS: set[str] = set()
 _BRAINS_LAST_GOOD: list[BrainSummary] | None = None
 _BRAINS_REFRESHING = False
 _BRAINS_REFRESH_TASKS: set[asyncio.Task[None]] = set()
+_GRADE_REFRESH_TASKS: set[asyncio.Task[Any]] = set()
 
 # Brain-wide uncertainty overview is bounded/cheap, but cached per brain for a short
 # window anyway to keep the dashboard's polling off the storage hot path (same
@@ -115,6 +116,10 @@ def _schedule_grade_refresh(storage: NeuralStorage, brain_name: str, key: str) -
         return report
 
     task = asyncio.get_running_loop().create_task(_refresh())
+    # Strong reference: the event loop discards tasks nobody holds, exactly as
+    # the brain refresh above already guards against.
+    _GRADE_REFRESH_TASKS.add(task)
+    task.add_done_callback(_GRADE_REFRESH_TASKS.discard)
     task.add_done_callback(_done)
 
 
--- a/tests/unit/test_dashboard_brains_scope.py
+++ b/tests/unit/test_dashboard_brains_scope.py
@@ -32,7 +32,9 @@ Two layers of coverage per endpoint, mirroring test_route_hub.py:
 
 from __future__ import annotations
 
+import asyncio
 import os
+from collections.abc import AsyncIterator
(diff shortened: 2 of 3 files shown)


acidkill (approved):
Reviewed the full diff. Strong task references mirror the brains-refresh pattern, and the live-test drain keeps teardown on the owning loop — addresses the 120 s teardown hang (#224). Green CI.

